PR #1, opened 2017-02-22
Pull Request for CPU, SCSI, VDFS, 32016

SteveFosdick:
This is everything I believe is complete.  By all means query anything that is not clear.

I have removed -Werror for the time being.  It was that or disable warnings for unused variables which I think are rather useful.  The problem is with the NS32016 NDis.c file.  Normally I would resolve the unused variable issues but as sometimes the variable is set to the value returned from a function, and the function may have side-effects, I think best to check with hoglet before doing that.


fordp2002:
I wrote NDis.c I can look to fix it and send a pull if you like.

ThomasAdam:
@fordp2002 -- yes please.  :)

fordp2002:
It would be easier to fix this once it is actually included in the startdot repository. It is just a warning and I am more than happy to fix the warnings. I did all the major architectural changes to the 32016 SoftCore. Hoglet had a major input in writing code to implement a lot of the op codes and lots of fixes, debugging and testing.

ThomasAdam:
@SteveFosdick -- thanks, I was just taking a look at this, but I cannot compile the net result of your pull-request.  I'm getting a tonne of errors in `src/fdi2raw.c` -- it's because the macros used when `_DEBUG` is not defined aren't doing the right thing.  Here's an example:

```
fdi2raw.c: In function 's0a':                                                       
fdi2raw.c:492:81: error: expected expression before ')' token                    
  bem_debugf("s0a:bits=%d,data=%s", bits, datalog(fdi->track_src, (bits + 7) / 8));
                                                                                 ^
fdi2raw.c: In function 's0b':                                                                                          
fdi2raw.c:512:82: error: expected expression before ')' token                            
  bem_debugf ("s0b:bits=%d,data=%s", bits, datalog(fdi->track_src, (bits + 7) / 8));
                                                                                  ^                                                                                         
fdi2raw.c: In function 's0c':                                                                                                                                                                                                               
fdi2raw.c:533:82: error: expected expression before ')' token                                             
  bem_debugf ("s0c:bits=%d,data=%s", bits, datalog(fdi->track_src, (bits + 7) / 8));                 
```

Indeed, just `#define _DEBUG` at the top of that file to see how far it gets results in other compilation errors.

Can you take a look for me?  Thanks.  Is this related to your changes or @hoglet67?

SteveFosdick:
Thomas, 

This is a bit of a weird one.  I think I can see what is going on, to an extent.  When _DEBUG is defined datalog should be a real function and so should bem_debug.  When _DEBUG is not defined datalog is a macro that expands to nothing so after the inner macro expansion this would make the first line:

bem_debugf("s0a:bits=%d,data=%s", bits, );

What this does not explain is why this compiles when I try it in the context of the master from my repository.  I don't know if this a gcc version thing or whether your file, after merging, has some subtle difference.

Could you post the errors from compiling with _DEBUG defined, please,as again that works fine for me?

SteveFosdick:
Ah, having just finished writing that I noticed you mentioned putting #define _DEBUG just at the top of the fdi2raw.c file.

When _DEBUG is defined for the project as a whole there is a real function called bem_debugf.  When _DEBUG is not defined it is a macro call that expands to nothing.  If you define _DEBUG for just this one file the compiler will generate a call for a function that doesn't exist so I'd expect a link failure and maybe a warning about not having seen a prototype for that function.

ThomasAdam:
I haven't done any merging yet, so this is just me checking out your repository and trying to compile from the `master` branch.

Without any modification to the code at this point, it's still not compiling for me.  The buildlog is attached.

Furthermore though, I'd like us to discuss your use of `alloca()`.  As you might know, FreeBSD doesn't ship alloca.h which means there's a portability problem.  Moreover though, might I recommend the use of `asprintf()` in those cases?

[buildlog.txt](https://github.com/stardot/b-em/files/797956/buildlog.txt)

SteveFosdick:
I'll look into the alloca.  I think asprintf was not available on Windows, though.

In the meantime I can't reproduce the error.  Even cloning my repository into a clean directory from github , so there can be no cruft from a previous run, and running autogen.sh configure and make in order does result in any error message - build log attached.

That means I can't easily make a change and test if it has fixed it
[buildlog2.txt](https://github.com/stardot/b-em/files/797994/buildlog2.txt)
.  I know what I expect would fix it, though, which is to change line 75 of fdi2raw.c from:

#define datalog(src, len)

to:

#define datalog(src, len) (char *)0

if that does solve it I'll commit it back.  That variant compiles fine here.

ThomasAdam:
It helps.  :)

See this diff here for further fixes -- it's scoping problems with _DEBUG.  I'm not sure I particularly like this smattering of _DEBUG stuff to be honest.

```
diff --git a/src/fdi2raw.c b/src/fdi2raw.c                 
index 98381ca..e637c25 100644                              
--- a/src/fdi2raw.c                                        
+++ b/src/fdi2raw.c                                        
@@ -72,11 +72,11 @@ static char *datalog(uae_u8 *src, int len)                                                         
        return buf + offset2;                              
 }                                                         
 #else                                                     
-#define datalog(src, len)                                 
+#define datalog(src, len) NULL                            
 #endif                                                    
                                                           
-#ifdef _DEBUG                                             
 static int fdi_allocated;                                 
+#ifdef _DEBUG                                             
 static void fdi_free (void *p)                            
 {                                                         
        int size;                                          
@@ -1308,9 +1308,9 @@ static void fix_mfm_sync (FDI *fdi)  
                                                           
 static int handle_sectors_described_track (FDI *fdi)      
 {                                                         
+       uae_u8 *start_src = fdi->track_src;                
 #ifdef _DEBUG                                             
        int oldout;                                        
-       uae_u8 *start_src = fdi->track_src;                
 #endif                                                    
        fdi->encoding_type = *fdi->track_src++;            
        fdi->index_offset = get_u32(fdi->track_src);
```

ThomasAdam:
I'll add an implementation of `asprintf` in cases where it's not defined -- watch `master` for that.  This will then enable you to make those changes as well to alloca.

SteveFosdick:
Ok, the alloca stuff is done and I have committed it to master.  I am sure asprintf will come in handy in the future, though.

If I add your other two changes, i.e. moving the declarations of  static int fdi_allocated and uae_u8 *start_src outside the scope of the #ifdef _DEBUG I get warnings instead:

make (in directory: /home/fozzy/Computing/BBC/b-em/master/src)
  CC       b_em-fdi2raw.o
fdi2raw.c: In function ‘handle_sectors_described_track’:
fdi2raw.c:1312:10: warning: unused variable ‘start_src’ [-Wunused-variable]
  uae_u8 *start_src = fdi->track_src;
          ^~~~~~~~~
At top level:
fdi2raw.c:78:12: warning: ‘fdi_allocated’ defined but not used [-Wunused-variable]
 static int fdi_allocated;
            ^~~~~~~~~~~~~
  CXXLD    b-em
cp b-em ..

ThomasAdam:
Hmm -- I wonder why you're not seeing the same things I am in your environment.

Thanks though for those changes -- did you `git push` them, only I don't see them in your repository here yet.

I've committed a change to master on `stardot/b-em` which adds an `asprintf` compat layer, and tested that it works well enough by explicitly using it -- I'm assuming windows will cope with the implementation I've written.

SteveFosdick:
Forgot to push it.  I have now.

I'll have to come back to the other issue.  I'm assuming you are using a different version of gcc from me and, depending on how clever the control flow analyser is, it does or does not detect all the unreachable code.

ThomasAdam:
Interestingly, if I use clang rather than GCC, then everything seems to compile.  However, your use of `tdestroy` from search.h is non-standard.  It won't work on BSD or Windows.  Can you add a wrapper for `tdestroy` or use something which isn't a GNU extension?

Also, please don't forget about our `x_fopen` wrapper.

ThomasAdam:
So I've taken everything from here so far, rebased the whole lot on top of master and pushed the result here on this branch:

https://github.com/stardot/b-em/tree/ta/stevefosdick-fixes

(That's `ta/stevefosdick-fixes`).  Note that the end result is car-crash of a mess because I'm also trying to compile in 32016 support from these commits which isn't complete, it seems.  Even disabling that results in linker problems with vasfs functions not being found.

Any ideas welcomed.  :)  I'm wondering if I need to back out the NS32016 support, and let @hoglet67 deal with it after this lot is through review?

hoglet67:
Thomas and Steve, I'm happy to re-prepare the 32016 changes after this merge is done. The NS32016 in Steve's pull request should be complete, it just contains some bugs.

I think the problem is that the files are missing from the new build system.

Take a look at this commit for what I needed to add to get things building:
https://github.com/hoglet67/b-em/commit/3efea6330d3fd91f9278166a527294ba09d93c2b

Dave

SteveFosdick:
Looking at that diff I did need to do something like that to get it to compile as Thomas's conversion to GNU autotools was done with the previous version of 32016, i.e. before I merged the working 32016 changes from you which put 32016 code in a subdirectory.

It is probably best to wait for the time being - I am resolving a couple of portability niggles.  if we can get this to a point where it compiles reliable for different people and thus commit the merge it will be easier to make further changes.

SteveFosdick:
Thomas,

I think I have got to the bottom of why we got different results around calls to bem_debugf.  It is this code here from b-em.h:
```
`// Remove debugging calls if debug not selected.
#ifndef _DEBUG
#define bem_debug(s) {}
#if __STDC_VERSION__ >= 199901L
#define bem_debugf(format, ...) {}
#endif
#define debug_open()  {}
#define debug_close() {}
#endif`
```

So this is trying to do a fairly standard thing - have some code that "just disappears" if the executable is not being built with debug enabled by doing again, a fairly standard trick of defining a macro that overrides a function name and expands to nothing so the function concerned is not called.

The reason for the version test of __STC_VERSION__ is that bem_debugf is a printf-style function that takes a variable number of arguments and before C99 variable argument macros were a GNU extension.  If variable argument macros are not available it should "gracefully degrade" to a function call to an empty function.  The difference, of course, is that when the compiler sees the empty macro expansion it can work out that some variables whose values are only used to print debug messages are not needed at all and generates the warning for an unused variable.  If the compiler is generating the function call, because the function is in a different module, the compiler still has to generate the call for it as it doesn't know what that function does.  That means if I try to eliminate the warning by making the declaration of the "unused" variable conditional on _DEBUG being defined it works fine on a C99 compiler which is using the empty macro expansion but fails on a compile that is generating the function call because the function call does need the variable defined.

One possible way out of that is to change that bit of header so that if debug in enabled in generates a declaration of an external function whereas if debug is off it generates a definition of an empty 'static inline' function.  This means the two cases should parse the same but in the case debug is enabled a decent compiler can optimise away the empty function.

ThomasAdam:
@SteveFosdick - Thanks, that's a good call indeed.  IIRC, there's `__printflike__` which we could probably get away with using here, no?

SteveFosdick:
Is __printflike__ and additional attribute you can attach to a function to hint the compiler that it should check the consistency between the format string and the actual arguments as if this was printf?  If so, yes, there is one of those functions in each of the error, warn and debug pairs.

I'll try a modification to implement and let's see if that results in something that compiles more reliably.  I'll let you know when I have committed it.

SteveFosdick:
Ok, the latest master uses the static inline function technique and also sets the printf format attribute of the functions that printf-style format strings.  The resulting warnings included some bugs so I have fixed those too.

SteveFosdick:
NS32016 warnings now fixed too.  These were the same cause as the bem_debugf one (though a different macro from a different source it was the same pattern).

ThomasAdam (commented):
Ah, there's overlap here, Steve.  As part of my review of all of this, I've already added a compat layer for this, but it did differently.

SteveFosdick:
And I hope tdestroy now fixed too in the latest master.

Obviously when the tsearch and similar functions were first included in the Unix standard C library the assumption was that programs were small and short lived so there would be no need to destory a search tree and free the memory for re-use and it seems that is what got standardised.

Unfortunately if a particular standard library's tsearch implementation lacks the tdestory we can't use it as we need it to avoid a memory leak.  In the event tdestroy is missing, the tsearch functions from the GNU C library are compiled into B-em instead to resolve the issue.  It is not possible to provide just the missing function as this needs to traverse a data structure built by the other functions and it has no knowledge of how other C libraries implement these trees.

SteveFosdick:
Thomas, for what?  For tdestory, the debug/non-debug of the 32016 code?  Where can I see your version?

ThomasAdam:
@SteveFosdick - for adding in a check for `tdestroy`.  It's fine, it's just duplicated effort -- I hadn't appreciated you'd also fixed it.

SteveFosdick:
I was going to ask you how to check a function in GNU autotools which I have always regarded as a black art but it turned out to be simple.  What solution did you come up with to provide the missing function?

ThomasAdam:
@SteveFosdick -- I've pushed a branch to `stardot/b-em` called `sf-master`.  This contains all of your original commits listed in this pull-request, PLUS some stuff from me.  Can you please check that functionally it matches what you expect?  In doing this work, I've had to resolve a few merge conflicts, and there's no guarantee I've done that 100%.  Also, I've removed `-Werror` again, until such time that @hoglet67 sends over some changes for 32016 emulation once all this is through.

I seem to also get these printed to stderr:

```
scsi lun 0: unable to open data file scsi/scsi0.dat: No such file or directory                                         
scsi lun 1: unable to open data file scsi/scsi1.dat: No such file or directory                                         
scsi lun 2: unable to open data file scsi/scsi2.dat: No such file or directory                                         
scsi lun 3: unable to open data file scsi/scsi3.dat: No such file or directory                                         
```

Are there files left to include?

SteveFosdick:
Those scsi warnings are because I ported the SCSI module from BeebEm across and it is looking for disk image files for four hard discs.  It doesn't really make sense to ship these because unlike floppies there isn't really a standard size of hard disk - people may want to devote more or less space to one.  if you were to turn of SCSI from the Disc menu they should go away.

I'll checkout your merged branch and see how it goes.

ThomasAdam:
Oh, and as for `search.h` -- yeah, it'll do for now, but you know that TAILQ and friends from `tree.h` are likely on their way, right?

SteveFosdick:
TAILQ?  I used tsearch because I wanted a data structure I could stick filenames into in the order the come from the hosts readdir call (which is effectively random) and return them in sorted order to the BBC and this seemed neat solution because it also naturally allocates dynamically rather than have to have a dynamically expanding array and then sort it at the end.

SteveFosdick:
Missleading indentation warnings in 6502.c again.

ThomasAdam:
I probably chose the wrong hunk when fixing the conflict -- hopefully that's the easier one(s) to solve.

SteveFosdick:
The SBC fix for the 6502 has also reverted.

ThomasAdam:
If it turns out there's significant portions of things where I've chosen the *wrong* side of the conflict, let me know.  But if you're OK to fix them up, please do so and I'll sort the mess out when you're done.

SteveFosdick:
Something ocurred to me: when I merged my original repository into a clone of your stardot one the terminology used by emacs smerge mode was exactly backwards, i.e. the option labelled "mine" was actually the incoming change and "other" was my change.  I assume that was becuse it was working in the context of the clone of the stardot repository.  Might you have chosen the exact opposite of what you intended in every case?

ThomasAdam:
@SteveFosdick -- no, I've not done that.

SteveFosdick:
So which files have you made functional changes to, i.e. which one can I not resolve by simply putting back the file I had working?

SteveFosdick:
Looking through the source tree it seems that every single change I have made had been reverted.

ThomasAdam:
Are you looking at the right thing here?  Because what I'm looking at has the SBC changes in per your original set.

SteveFosdick:
Maybe not.  I assumed when I chose a branch in github and then used their clone URL it would clone the branch.  It seems maybe it didn't do that.  How do I get to the right branch?

ThomasAdam:
@SteveFosdick -- I've opened up a pull-request (temporarily) here:  https://github.com/stardot/b-em/pull/3

If you want to check locally then from your checkout of the `stardot/b-em` repo:

```
git fetch
git checkout -t origin/sf-master
```

That'll put you on the branch in question.

SteveFosdick:
That's looking better already.  I'll try some more tests.

ThomasAdam:
@SteveFosdick -- please pay attention to the GUI elements you changed for the SCSI emulation -- check I haven't fudged that area since it was a bit niggly to resolve merges there.

SteveFosdick:
GUI looks good.  CPU passes the test of starting the BCPL ROM.  I'll have to look for the proper CPU test suite but promising so far.

ThomasAdam:
@SteveFosdick -- just noticed, if I select a Model B variant to emulate, then *disable* VADFS from the menu, and then SHIFT+BREAK or some variant thereof, then I am left with a broken emulator.  See screenshot attached.

bem1

If I then press SHIFT + BREAK again, I get this:

bem2

SteveFosdick:
If you have SCSI enabled ADFS expects to use it and drives 0-3 are the hard disks and drives 4 and 5 are the floppies so that could be what is going on.  I'll try it.

ThomasAdam:
Ah -- I see.  Yes, unchecking SCSI as a harddisk option fixes this.

SteveFosdick:
You also get errors from ADFS if you have a DFS floppy in the drive.

ThomasAdam:
This is encouraging though because now we're into functional testing... ;)

SteveFosdick:
Maybe the default should be VDFS off and hard disc none so there are no surprises.

ThomasAdam:
@SteveFosdick - Yeah, OK.  Want me to do that?

SteveFosdick:
Ok.

SteveFosdick:
Both main and tube CPUs passed the 6502 functional test suite.

SteveFosdick:
It seems the 32016 is broken, though, and it did work.  Now all I get is a flashing cursor if I enable it.

ThomasAdam:
@SteveFosdick -- given @hoglet67 has further changes, can we defer that?
@SteveFosdick  -- also, can you please make the GUI changes?  I'm off to bed now...

SteveFosdick:
I'll see if I can do it tomorrow, the GUI change that is, you're right it's late.  This is progress, though.

ThomasAdam:
Very good progress -- thanks for your hard work and patience thus far!  :)

SteveFosdick:
Ok, I have fixes for both the 32016 and the GUI change.  I have committed them to my local copy of the sf-master branch but it won't let me push it back.  How do I get the changes to you?

ThomasAdam:
@SteveFosdick -- I've invited you to become a member -- if you accept that, you will then be able to push to that branch.

SteveFosdick:
Ok, I have pushed those changes, all trivial.  I need to head out now but will check back some time in the afternoon.

ThomasAdam:
@SteveFosdick -- thank you, that looks good.  I think on this particular topic, we're done.  I'll merge that branch into master. I'm closing this now.

Fantastic work, @SteveFosdick.  Thanks once again for all your help!

PR #2, opened 2017-02-23
Update git ignore (test of github pull request)

fordp2002:
Trivial change to test pull request.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,6 @@ install-sh
 missing
 tags
 src/Debug/
+*.opensdf
+*.sdf
+*.sln
--- a/.gitignore
+++ b/.gitignore
@@ -17,7 +17,10 @@ etc/
 install-sh
 missing
 tags
+
+# Windows Visual Studio Files
 src/Debug/
+src/Release/
 *.opensdf
 *.sdf
 *.sln
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,4 @@ etc/
 install-sh
 missing
 tags
+src/Debug/
--- a/src/win.c
+++ b/src/win.c
@@ -952,4 +952,9 @@ LRESULT CALLBACK WindowProcedure (HWND hwnd, UINT message, WPARAM wParam, LPARAM
         }
         return 0;
 }
+
+FILE *x_fopen(const char * file, const char * mode)
+{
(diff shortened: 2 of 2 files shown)


ThomasAdam:
Thanks.  Can you please ensure you keep the file sorted, please?  I've found that to be very handy for this file, in particular.

fordp2002:
Well I added the ignore from the SourceTree gui. I will go in and edit it in future I prefer the execeptions stored under a heading too.

fordp2002:
I have added a heading now

fordp2002:
Yes I should have said Visual Studio 2013 Community Edition. I have already said I would also build using the open source toolchain if I had the instructions. I already started two wiki entries. I may even try BSD for the hell of it I have so many spare PCs lying about ;)

ThomasAdam:
What's happening with this pull-request?  If anything?

fordp2002:
Sorry I have been away. I will try and get some time to get B-Em building on Visual Studio. The only change I was proposing was Visual Studio exclusions. I am still not clear how to handle line endings at the moment, I will get to the bottom of that at some point.

SteveFosdick:
With regard to the line endings, the git repository uses LF only, i.e. the Unix convention.

Although many Windows programs only work with the Windows CR,LF convention editors for programmers can usually handle Unix format and maybe MAC (and Acorn) format too (CR-only) so maybe Visual Studio will too.  Even so I think there are some git options to set here - see https://git-scm.com/docs/gitattributes - to have line endings normalised as files are committed because I am sure, even if Visual Studio can handle Unix line endings, for new files it is bound to use the Windows convention and then we end up with a mixed bag.

hoglet67:
@SteveFosdick -- don't we actually already enable the git auto line ending magic?
https://github.com/stardot/b-em/blob/master/.gitattributes

@fordp2002 -- if you clone the repository on Windows, then any "text" fines should be converted to crlf endings. Then when you check back in, they are converted back again. At least, that's how I think it's meant to work....

fordp2002:
Converting is one way of working. The other is just to use Unix line endings on Windows which is the way I would prefer to work as I think it is best practice. Visual Studio is happy with Unix Line Endings even if notepad is not! However automatically created files have to configured to retain there settings.

PR #3, opened 2017-02-25
sf-master

ThomasAdam:


hoglet67:
Great work Guys.

My plan (when you give me the go ahead to start) is to rename my old b-em repository, and then refork from startdot/b-em and then prepare a pull request for the latest 32016 emulation bug fixes.

One challenge I had is trying to keep this code in sync with PiTubeDirect. I see a bunch of white spaces changes have crept in with this recent burst of work. What white space policy do you have set up? (it looks like auto deletion of trailing while space plus converting tabs to space). I'm happy to adopt what ever, and then push these white space changes back to PiTubeDirect.

One more point:

To use the 32016 emulation code in both PiTubeDirect and B-Em there is a little bit of conditional compilation (two places) with the symbol -DBEEBEM (no idea why it ended up being called that!). I see some of this has been removed. Are you averse to me putting it back? Is there a symbol already defined in the new build system that I could hijack to identify compilation in B-Em?

I've got some time to start this morning.

Is the state of master reasonable for me to use as a starting point?

SteveFosdick:
David,

I can't speak for Thomas but for modules I don't expect to actively maintain I am not bothered if the white space is at odds with the rest of the project. The changes you see maybe because I have used an editor that canonicalises white space automatically. If you want to just overwrite the 32016 files with the latest ones that's fine by me.

On the conditional compilation I suspect Thomas didn't know this is code is shared with another project that is still actively developed and therefore tried to make it "B-em native". That broke it completely and the latest change I pushed was to put the conditional compilation back. The one difference is I added #define BEEBEM rather than try to find the right place to put -DBEEEBEM in the new autotools make system but that was just an expedient - you're welcome to change that too.

SteveFosdick:
David,

Another thought on the conditional compilation, given how easy I found to port the SCSI module from BeebEm to B-Em I wonder if we might share modules between the two at some point in the future. Maybe even this 32016 code, though I am not suggesting you drive that. It may be worth choosing a more logical name if the symbol is to be defined for the whole project.

fordp2002:
There a licensing issues with the 32016 code as that is GPL licenced. I respect this so it should not go in to Beebme as that is not compatible as far as I know.

ThomasAdam:
b-em is GPL -- so no no conflicts there.  What are you thinking of, @fordp2002?

fordp2002:
BeebEm is not GPL that is what I was thinking of!

ThomasAdam:
@fordp2002 --  According to the Windows "version" [0]:  http://www.mkw.me.uk/beebem/index.html it is:  http://www.mkw.me.uk/beebem/BeebEm414.zip

The UNIX port is different, it seems.  Might be worth getting in touch with David Gilbert to find out what's going on there.  Since he claims to have based this off the Windows version which is GPL, he's in violation of that in changing the copyright to something which reads anti-GPL.  I don't think he has any basis to do this, and hence his UNIX port is GPL by definition.

[0] This is a bloody mess...

hoglet67:
According to the [Change Log](http://www.mkw.me.uk/beebem/CHANGES.txt), Mike Wyatt changed the licensing to GPL in version 4.10 (July 2009). So it's possible David Gilbert forked before then.

http://stardot.org.uk/forums/viewtopic.php?t=2445

fordp2002:
My impression was that BeebM is not GPL. If this is correct it is a major issue. If I am wrong I am sorry but that was my impression. B-Em IS gpl hence me spending my time on it. P.S. Still on Holiday, ;)

PR #8, opened 2017-02-26
Mass reindentation with gindent(1)

ThomasAdam:
In order for b-em to remain consistent with more than one person working on
it, it's important to have consistency within the code.

The suggestion is to use GNU Indent (gindent(1)) to do this -- and the Linux
kernel style seems appropriate here, hence:

    for i in src/*.[ch]; do echo "$i" && gindent -linux -l78 -di16 "$i"; done

Eventually, this could be automated, but this seems like the easiest set of
rules to standardise on.

Note that this deliberately just targets the code under src/ and not anything
else -- as this is core b-em, and I don't want to touch other third-party code
we also choose to include as part of b-em.


hoglet67:
Thomas, I would suggest delaying this until we are in a stable state with no work "inflight". I have some more changes stacked up, and doing this now will just cause unnecessary work.

I'm please to see you have avoided the NS32016 sub-folder. This is shared with another project with it's own white space constraints, and I would prefer it to remain as is.

hoglet67 (commented):
Thomas, I would suggest delaying this until we are in a stable state with no work "inflight". I have some more changes stacked up, and doing this now will just cause unnecessary work.

I'm pleased to see you have avoided the NS32016 sub-folder. This is shared with another project with it's own white space constraints, and I would prefer it to remain as is.

ThomasAdam:
Oh, absolutely, @hoglet67.  I'm merely leaving it here to float the idea in peoples' minds.  That's why I wanted feedback more than anything.

chrisn:
This kind of change is almost never a good idea. It introduces an enormous diff, makes tracing changes to previous versions next to impossible, all for no substantial improvement to the codebase.

SteveFosdick:
@hoglet67 Do you have a defined style for NS32016?  I know these can become religious wars but I am just thinking that if you already have a style, and if Thomas is set of consistency, we could both use the same style.

SteveFosdick:
Thinking more generally about this I certainly share the annoyance of opening code up in an editor and finding the structure of the code is not clear because the code uses white space for indentation in a different way from my editor's defaults.  Whether reformatting is the answer I am not so sure.  Even if the code was reformatted there's still a chance that someone's editor will do the wrong thing.  Maybe each file should use a consistent style and include one of those "label" lines (an emacs invention IIRC) to declare what that convention is so someone working on it can:

a) have it displayed correctly.
b) edit consistent with that style.

I not convinced on the benefit of having a tool force line breaks into the code.  Yes, it means you can use the file on a 80x24 terminal but do many people use of these today?  If the intention of limiting line length is actually to limit nesting depth an automated tool will not help.  I also don't see the benefit of adjusting the white space within a line.  I can't get excited about whether there is a space around an equals sign or after a comma etc.

On the question of being able to see changes across a mass reformat I wonder how many diff tools have the option to ignore white space changes, i.e. consider any pair of lines where the only changes have been white space to be the same.  GNU diff does and this provides the way to be able to see what has functionally changed even across such a reformat.

hoglet67:
The more I think about this, the more I'm tending to agree with @chrisn 

@SteveFosdick Regarding the 32016 code, whilst we don't have rigid constraints, one of the main developers of the 32016 (fordp/Simon Elwood) needs plenty of additional white space for C code to be readable to him (I think he has a condition similar to dyslexia). He's put a lot of time into manually reformatting this code. So I do feel strongly that the current white space formatting needs to stay.

Personally I'm less concerned, but I do have a strong preference for spaces over tabs.

fordp2002:
I do struggle with densely packed code due to word blindness. We should think of the Copro's as separate projects in my view as they can be shared with the PiDirect and SPI tube projects.

ThomasAdam:
@fordp2002 @hoglet67 -- I do have the option of using submodules here for such a purpose -- this would mean it's kept separate from b-em and pulled in when needed from a single source, rather than us having to copy the code in each time.

SteveFosdick:
This idea of shared copros is interesting.  I know for the 32016 there is a little bit of conditional compilation used to make the same source code suitable for both environments.  Presumably we could do something similar with the other copros in B-em and these could be ported back the other way to run on the Pi.

I take it any copro will just communicate though the tube interface, either real in the case of the PiDirect or emulated in the case of B-Em and shouldn't need to communicate in any way with the other modules in B-Em that emulate hardware attached to the I/O processor?  I am just wondering what, if anything, could be done to make sharing modules like that easier.  What would a soft copro expect its host environment to provide?  Some means to get/set config, write to a log etc?  Some means of controlling the emulated clock speed and, in the case of B-Em, maintaining this as a multiple of the I/O processor clock speed?  I am just wondering if this can be standardised to the point all the copros work in either project.

ThomasAdam:
@SteveFosdick - simpler than that, I'm referring at `git clone` time, etc., to pull in the latest 32016 code -- it's a common technique for submodules.

SteveFosdick:
@ThomasAdam I think that is addressing a slightly different question.  You're answering "how to we collect some modules together at the point we wish to compile the project" and I am talking about what changes we need to make to the modules themselves so the same code compiles from within both contexts, i.e. doesn't reference header files that don't exist, call functions that don't exist or do nothing because the host environment didn't call its run() function.  At the moment 32016 is the only shared module but this is about extending that to others.

ThomasAdam:
Ah, OK.  Understood.

fordp2002:
I would suggest adding an .editorconfig file and enforcing a code style going forward. I am happy with spaces or tabs and and most tab sizes from 2 to 4. The file above should automatically bring things inline for editors that support it. I do not have strong feelings on a single reformat. I can see the merits on both sides.

ThomasAdam:
Possibly -- editorconfig is much less likely to reach all people than using modelines at the top of files, IMO.  But to get to the point of using either, you have to decide what that format should be, and the last thing I'd want is a complete mixture of old and new(er) code formatted differently via vicarious means.

fordp2002:
The point about my suggestion is as code is changed it starts to normalize and in time we get where we want with less no need for a big bang discontentuity.

ThomasAdam:
Since this is like asking people if they take sugar in their tea/coffee -- you can all go cold turkey and suck this one up.

Don't say I didn't warn you, and don't dare moan when surrounding code which will have inherent white space changes happens.

You've been warned.

SteveFosdick:
Github suggests you have closed this without merging but you comment suggests you have applied it.

ThomasAdam:
Heavens, no.  I've not applied this.

PR #10, opened 2017-03-01
Reformatted Makefile.win to reduce conflicts

hoglet67:
Mostly cosmetic, but keeping files sorted is good, and should reduce merge conflicts.

### DIFF
--- a/src/Makefile.win
+++ b/src/Makefile.win
@@ -1,17 +1,98 @@
 VPATH = . resid-fp NS32016
+
 CPP  = g++.exe
 CC   = gcc.exe
 WINDRES = windres.exe
+
 CFLAGS = -O3 -DBEM -DVERSION=\"Win32\"
-OBJ = compat_wrappers.o 6502.o 6502tube.o 32016.o Decode.o mem32016.o Trap.o Profile.o NSDis.o 65816.o acia.o adc.o adf.o arm.o cmos.o compact_joystick.o compactcmos.o config.o csw.o ddnoise.o debugger.o disc.o fdi2raw.o fdi.o i8271.o ide.o keyboard.o logging.o main.o mem.o model.o mouse.o pal.o savestate.o scsi.o serial.o sn76489.o sound.o soundopenal.o ssd.o sysvia.o tape.o tapenoise.o tsearch.o tube.o uef.o uservia.o vdfs.o via.o vidalleg.o video.o wd1770.o win.o win-catalogue.o win-keydefine.o x86.o z80.o resid.o b-em.res
-SIDOBJ = convolve.o envelope.o extfilt.o filter.o pot.o sid.o voice.o wave6581__ST.o wave6581_P_T.o wave6581_PS_.o wave6581_PST.o wave8580__ST.o wave8580_P_T.o wave8580_PS_.o wave8580_PST.o wave.o
+
+OBJ = \
+    6502.o \
+    6502tube.o \
+    65816.o \
+    acia.o \
+    adc.o \
+    adf.o \
+    arm.o \
+    cmos.o \
+    compact_joystick.o \
+    compactcmos.o \
+    compat_wrappers.o \
+    config.o \
+    csw.o \
+    ddnoise.o \
+    debugger.o \
+    disc.o \
+    fdi2raw.o \
+    fdi.o \
+    i8271.o \
+    ide.o \
+    keyboard.o \
+    logging.o \
+    main.o \
+    mem.o \
+    model.o \
+    mouse.o \
+    pal.o \
+    savestate.o \
+    scsi.o \
+    serial.o \
+    sn76489.o \
+    sound.o \
+    soundopenal.o \
+    ssd.o \
+    sysvia.o \
+    tape.o \
+    tapenoise.o \
+    tsearch.o \
+    tube.o \
+    uef.o \
+    uservia.o \
+    vdfs.o \
+    via.o \
+    vidalleg.o \
+    video.o \
+    wd1770.o \
+    win.o \
+    win-catalogue.o \
+    win-keydefine.o \
+    x86.o \
+    z80.o \
+    resid.o \
+    b-em.res
+
+NS32KOBJ = \
+    32016.o \
+    Decode.o \
+    mem32016.o \
+    Trap.o \
+    Profile.o \
+    NSDis.o
+
+SIDOBJ = \
+    convolve.o \
+    envelope.o \
+    extfilt.o \
+    filter.o \
+    pot.o \
+    sid.o \
+    voice.o \
+    wave6581__ST.o \
+    wave6581_P_T.o \
+    wave6581_PS_.o \
+    wave6581_PST.o \
+    wave8580__ST.o \
+    wave8580_P_T.o \
+    wave8580_PS_.o \
+    wave8580_PST.o \
+    wave.o
 
 LIBS =  -mwindows -lalleg -lz -lalut -lopenal32 -lgdi32 -lwinmm -lstdc++
 
 all : b-em.exe
 
-b-em.exe: $(OBJ) $(SIDOBJ)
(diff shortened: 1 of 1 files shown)


hoglet67:
Another small question: Is it bad form to issue small pull requests from Master?

If so, then I can create a feature branch and resubmit.

ThomasAdam (approved):
OK.

ThomasAdam:
@hoglet67 -- hopefully I've alluded to the answer in the `SUBMITTING_PATCHES.md` file -- but the way this works is that it's best to use topic branches for your own work.  Given that the `master` branch is the canonical working source of the code, putting your changes directly on that branch realistically makes your life harder when there's changes to that branch which aren't related to the feature you're developing.  Separating out the work you do from `master` means you can just roll-forward with changes to `master` without conflict.

hoglet67:
Understood, I'll use feature branches in future.

One thing I'm finding about the current workflow is that it seems to involve the unnecessary re-writing of commits.

The above commit  c06aad8 has now appeared in master as 119c333.

What this means is my own master branch has diverged (i.e. one commit being, one commit ahead). Even though there were no changes.

So this means I need to rebase all work I have done that has descended from that commit, i.e. the most recent 6 commits on:
https://github.com/hoglet67/b-em/commits/db-music5000-safe

This seems to be down to the committer changing:

```
David Banks@ul30hog MINGW64 /c/MinGW/b-em/src (db-music5000)
$ git show --format=full c06a > 1

David Banks@ul30hog MINGW64 /c/MinGW/b-em/src (db-music5000)
$ git show --format=full 119c > 2

David Banks@ul30hog MINGW64 /c/MinGW/b-em/src (db-music5000)
$ diff 1 2
1c1
< commit c06aad84d95baef77c1a1d4c6b8189240d8a1c99
---
> commit 119c33356adbe03706c742b1b1d87cd4970ab0f1
3c3
< Commit: David Banks <dave@hoglet.com>
---
> Commit: Thomas Adam <ThomasAdam@users.noreply.github.com>
```

Do you understand why this is happening?

It's making this workflow close to unworkable for me.

hoglet67:
This behaviour is at least described here:
https://help.github.com/articles/about-pull-request-merges/#rebase-and-merge-your-pull-request-commits

> The rebase and merge behavior on GitHub deviates slightly from git rebase. Rebase and merge on GitHub will always update the committer information and create new commit SHAs, whereas git rebase outside of GitHub does not change the committer information when the rebase happens on top of an ancestor commit. For more information about git rebase, see the "Git rebase" chapter from the Pro Git book.

Now why github think this is a good ideas is beyond me!

I've used more forced pushed in the last 2 days than I have in the previous 8 years of using git!

ThomasAdam:
That's right.

In your case, the work you've done on master will sit there -- which comes back to an earlier point about using Master as a tracking branch only.

I would suggest you do this:

```
git reset --hard origin/master
```

Then your problems go away.   Having looked at your copy of `master`, I see no additional commits which haven't already been merged.

hoglet67:
I understand how to reset a branch, but that's not really my point.

I suppose some of this is learning for me, as I've always used merge based workflows before (with lots of local-only rebase to avoid unnecessary merge commits).

My take away so far is:
- always prepare a pull request on a separate branch, which you then abandon (and delete)
- avoid basing local work on commits in that pull request, as they will be re-written

I appreciate there are benefits of trying to make development seem linear. But in reality it isn't, and retaining the history that shows shows these features were developed in parallel and then brought together is actually a good thing.

Take the Music-5000 work. Soon I would like to make the current work available upstream, in part because it's useful to get feedback, and in part because it's painful to keep rebasing. But I would like to continue work on the same branch, as there are still issues to resolve. It seems at present if I do a pull request from db-music5000 on hoglet67, I then have to stop work until that's merged. Otherwise, I end up being forced to rebase because github has updated the commit SHAs.

Now, maybe I've just been unlucky so far, in that every rebase has had conflicts, because there's been a fair amount of churn e.g. to the Makefiles, and some other overlaps.

I should say, I'm by no means new to git. having used it for about 8 years. What's new to me is github's network of forks and pull requests.

@SteveFosdick, I'm wondering is this just me, or are you also struggling a bit?

I do wonder if things would be simpler if the all the core developers were all working on branches in the upstream repository, rather than in forks. It removes one level of indirection.

ThomasAdam:
@hoglet67 -- so none of this in practise is painful, and what you've experienced is just unfortunate.  Given that I've explained to you why you should use topic branches, that's going to mitigate your concerns.

The other point here is that in using topic branches with the workflow I've been describing, it forces you to resolve any conflicts before the merge; that it becomes a fast-forward is a nice benefit through a side-effect.  If there's churn, that's a good thing.  In this case, there's been a lot of changes, so I'd expect that.

hoglet67:
> so none of this in practise is painful, and what you've experienced is just unfortunate. Given that I've explained to you why you should use topic branches, that's going to mitigate your concerns.

Are you always so patronising?

I'm trying to have a discussion about best practices.

Fine, then I'll spend my time else where....

SteveFosdick:
@hoglet67 Dave, I have found the whole process of collaborating via github more confusing and with greater risk of error than I had expected.  On the other hand I don't have your experience in using git - I have used it locally for a few years, because it doesn't need a local server, and I know people have been raving about it, but my day job is with a large company that is slow to change.  I have used others VCSs in the past, though.

The pull request with the changed commit IDs I put in was the result of an attempt to keep my own forked repository synced with the stardot one which, in the way I think github is supposed to work, should be a straightforward thing to do and it seems it is not.

The bit of the manual I read yesterday mentioned, I think, two IDs associated with a commit: an author and a committer (I think they were the terms).  Is this what github is doing when changing those commits, i.e. retaining the author by making the committer the person who did the merge?

Anyway, like you I had wondered whether maintained a forked version of the repository was extra work that was not worth the candle and whether topic branches in the stardot repository would be better.  As a test of my understanding of this, if we used topic branches, wouldn't rebasing become optional?  If one was to pull the master branch into the topic branch and commit the merge immediately before submitting a pull request (to merge the topic branch into master) the resulting merge should apply cleanly?

ThomasAdam:
@hoglet67 -- patronising?  No.  If you feel that way, I can only profusely apologise as that was never my intent.

If either yourself or @SteveFosdick need help with this, so be it.  I've always said I'm happy to do that, and I stand by that point.  I have just pushed to `master` a clarification about keeping your fork in sync with upstream.  I will continue to do everything I can to ensure you are both able to work with the code base here.  I am keen to point you towards the `SUBMITTING_PATCHES.md` file, if only because that's what I see as the canonical reference for new contributors, so if that is water-tight, we're laughing.

None of what I am writing up or proposing is alien in anyway -- these are standard practises which I am hopeful you'll both embrace over time with my help.  I hope I can't say fairer than that.

I hope you can appreciate and see where I'm coming from on this, and if nothing's clear, then I will repeat my point that you can contact me however you wish.

Kindly,
Thomas

hoglet67:
Thomas, sorry for that post earlier which I subsequently deleted. I was out of order. I had a bit of a frustrating afternoon rebasing and re-resolving what felt like the same conflicts about three times. Your message saying "so none of this in practise is painful" came just at the wrong time!

I do appreciate the effort you are putting in to help all of get further up the rather steep learning curve of git. I'll try to persevere with the recommended workflow for a while longer.

On a separate note, the audio on B-Em is currently pretty awful. I'll open a seperate issue to discuss this.

ThomasAdam:
@hoglet67 -- Not a problem.  As for the audio, I've been reading your thread on the forums. Currently, the audio libraries b-em uses are external.  I meant to ask you via the forums what issues you're experiencing, but since you're here, I'll ask... what's up?  :)

hoglet67:
@ThomasAdam -- I've put my thoughts on the audio issue here: https://github.com/stardot/b-em/issues/11

PR #13, opened 2017-03-03
Db music5000

hoglet67:
Thomas, here is a first pull request for the Music 5000 work. It's tested on both Windows and Linux, and sounds very close to the FPGA now.

Also included is bit of a clean up of the sound layer in general to:
- use #defined constants for buffer sizes and frequencies, rather than have dozens of identical literal values splattered all over the code
- reduce "cut and paste" of identical code for the different sound sources
- improve the commenting


ThomasAdam (changes requested):
This looks fine to me, @hoglet67.  There's a couple of things though which are worth fixing (see the diff below):

* Add `(void)` for function prototypes which expect no arguments -- let's try and be ANSI where we can, especially for new code.

* Add missing export for `music5000_reset()`

* Reindent some obviously misaligned code.

* For menu entries, I'm striving towards `D_CLOSE` rather than `D_O_K`.

Can you squash the following in to your pull-request and push the change out?

```
diff --git a/src/linux-gui.c b/src/linux-gui.c
index d752652..76a6335 100644
--- a/src/linux-gui.c
+++ b/src/linux-gui.c
@@ -662,7 +662,7 @@ int gui_music5000()
 {
         sound_music5000 = !sound_music5000;
         gui_update();
-        return D_O_K;
+        return D_CLOSE;
 }
 int gui_dac()
 {
diff --git a/src/linux-keydefine.c b/src/linux-keydefine.c
index 3ac8033..fa2fbad 100644
--- a/src/linux-keydefine.c
+++ b/src/linux-keydefine.c
@@ -326,6 +326,6 @@ int gui_keydefine()
                 d[x].y-=(d[0].h/2)-(256/2);
                 x++;
         }
-        return D_O_K;
+        return D_CLOSE;
 }
 #endif
diff --git a/src/music5000.c b/src/music5000.c
index e03aa83..9d7680d 100644
--- a/src/music5000.c
+++ b/src/music5000.c
@@ -66,9 +66,7 @@ static int disable;
 static short sam;
 static byte sign;
 
-
-
-void music5000_reset()
+void music5000_reset(void)
 {
 	memset(RAM, 0, 2048);
 	memset(phaseRAM, 0, 16 * sizeof(int));
@@ -81,9 +79,10 @@ void music5000_reset()
 	sam = 0;
 }
 
-void music5000_init()
+void music5000_init(void)
 {
-   int n;
+	int n;
+
 	for (n = 0; n < 128; n++) {
 		//12-bit antilog as per AM6070 datasheet
 		int S = n & 15, C = n >> 4;
@@ -95,11 +94,11 @@ void music5000_init()
 
 void music5000_write(uint16_t addr, uint8_t val)
 {
-   static uint8_t page = 0;
-   if (addr == 0xfcff) {
-      page = val;
-   } else if (((page & 0xf0) == 0x30) && ((addr & 0xff00) == 0xfd00)) {
-      addr = (page << 8) | (addr & 0xFF);
+	static uint8_t page = 0;
+	if (addr == 0xfcff) {
+		page = val;
+	} else if (((page & 0xf0) == 0x30) && ((addr & 0xff00) == 0xfd00)) {
+		addr = (page << 8) | (addr & 0xFF);
 		if ((addr & 0x0f00) == 0x0e00) {
 			//control RAM
 			RAM[I_WFTOP + (addr & 0xff)] = val;
@@ -115,7 +114,7 @@ void music5000_write(uint16_t addr, uint8_t val)
 	}
 }
 
-void music5000_update_6MHz()
+void music5000_update_6MHz(void)
 {
 	int c = (channel << 1) + (modulate ? 1 : 0);
 	switch (pc) {
diff --git a/src/music5000.h b/src/music5000.h
index 31fc04c..45724a4 100644
--- a/src/music5000.h
+++ b/src/music5000.h
@@ -1,3 +1,4 @@
-void music5000_init();
+void music5000_init(void);
 void music5000_fillbuf(int16_t *buffer, int len);
 void music5000_write(uint16_t addr, uint8_t val);
+void music5000_reset(void);
```

hoglet67:
Changes added, as a additional commit.

ThomasAdam (approved):

ThomasAdam:
Thanks, @hoglet67.  Looks good.

Is there anything I can do to make use of music5000?  I'm just curious.

SteveFosdick:
There is menu item to enable it under Misc/Sound and a disc under the discs directory that has some example tunes.

SteveFosdick:
Try autobooting the disc.

Once in AMPLE if you want to load an individual tune its "filename" LOAD RUN
backwards because AMPLE is an RPL (forth-like) language.

hoglet67:
Try following these steps:

- Set the system to be a Master 128 with no Co Processor.
- Enable Music 5000 in the settinngs/sound menu.
- Load discs/MUSIC5000.ssd into drive 0.
- Shift-Break to load the ROMs
- Shift-Break a second time to load AMPLE.
- Select "Load program" and then enter PPACH
- Select "Run program"
- turn the volume up!

Other tracks on that disk are PAVANE and CONCERT

ThomasAdam:
That's impressive indeed!  :)

hoglet67:
It really was an amazing piece of hardware. And the original was just built out of a bunch of TTL devices and a couple of RAMs. If you want to know a little more, Ed posted some links recently here:
http://forum.6502.org/viewtopic.php?f=4&t=4436

ThomasAdam:
Thanks, that's interesting.  I'm reminded of this episode:  https://www.youtube.com/watch?v=OF53Ykr1z6c&list=PLOtimvwAoYtlWZdX-MgnHrr-GUS7m8F7g&index=7

SteveFosdick:
@hoglet67 Dave, out of interest should this emulation handle the AMPLE written for the Music 500, i.e. the non Master one?  I think you said the hardware difference was minimal.

I found the Music 500 system disc on Colin Fraser's site at http://www.colinfraser.com/m5000/m5000.htm.  To start with that didn't work at all until I discovered the AMPLE file was all zeros.  Copying the AMPLE file from another disc seems to have solved that - it loads AMPLE and seems to run the programs but there are some oddities:

1. "Wild Boys" has some strange siren sounds.
2. "Pachelbel's Canon" is badly out of tune.
3. "Sex Crime" is also over the place pitch-wise.

Given the initial problem with this disc it is completely possible the issue is actually corrupt files, i.e. the AMPLE programs for those particular tunes, but as I can't find another copy of that system disc anywhere it is hard to compare.  Do you have one?

If not, any idea what may be going on?

hoglet67:
@SteveFosdick -- As far as I'm aware, the only hardware difference between the Music 500 and Music 5000 was one added resistor, to cope with a slight timing difference in the 1MHz bus on the Master compared to the Model B. So I would expect the original Music 500 disk to work. You could try the version here:
http://acorn.huininga.nl/pub/disks/originals/Ample/

SteveFosdick:
@hoglet67 Dave that seems to have the same problem with the AMPLE file being all zeros.

Looking in more detail at Pachelbel's canon I don't think it can be corruption.  The part that stands out as being out of tune is the bassline and all that does is play one bar of  DbBf, one bar of GdGA and repeat that 56 times.

Here's a screenshot of the piece parred down to just that bass part and it sound horrid:

screenshot from 2017-03-05 22-19-53

SteveFosdick:
So the problem appears to be pitch envelopes 4 and 5.  if I change them both to 1 it sounds fine.

hoglet67:
There do seem to be several versions of AMPLE NUCLEUS 1.00 floating around that are different. I wonder if the ROM contains some preset configurations that maybe changed over time.

SteveFosdick:
This is the disc-based AMPLE but I did only replace the AMPLE file which is short, i.e. seems to be some kind of bootstrap loader.  The main code is in a file called IMAGE and I could see if the other disc I have has a different version of that.  There is also a file called PRESET1 which may be the culprit.

hoglet67:
@SteveFosdick -- a couple of useful links I just found:

Music 500 User Guide
http://chrisacorns.computinghistory.org.uk/docs/Acorn/Manuals/Acorn_Music500UG.pdf

Original Tape Image:
http://www.stairwaytohell.com/bbc/archive/tapeimages/hardware/Hybrid/Music500_B.zip

I was able to follow the instructions and load AMPLE, then LOAD/RUN POPCORN

SteveFosdick:
Ok, I'll try that. Swapping IMAGE or PRESET1 didn't solve it.

SteveFosdick:
Ok, following the tape to disc process in the manual:

POPCORN - fine.
TUBELLS - fine.
TLORDS - fine
HANDOWN - fine
MAPLE - FINE
BRANDEN - horrid
SYNCRON - definitely not right.

Pitch envelope 4 is fine when AMPLE is first initialised but after having been setup by PRESET1 the variation in pitch seems much too wide to be musical.

hoglet67:
I also couldn't get BRANDEN to play correctly, even from tape. It is possible I guess we have a subtle emulation bug that affects a feature only used in that track. 

It would be useful to see what the hardware implementation(s) makes of this. I'll can try the FPGA this morning.

The other thing I was going to do was work out exactly what the Music 500 -> Music 5000 modification was doing. There are some pictures here:
http://stardot.org.uk/forums/viewtopic.php?p=15326#p15326

hoglet67:
FYI, It seems Music 500 AMPLE programs were not necessarily compatible with the Music 5000 AMPLE Nucleus ROM. Someone has produced a compatibility module:
http://knackered.org/angus/beeb/mu500.html

SteveFosdick:
That's interesting.  Yes, I noticed the incompatibility and there are some notes in volume 3 of the  Amplinex magazine (https://sites.google.com/site/amplinex003/features/ample-bce-to-ample-nucleus-conversion) on how to port your ample programs from the BCE (Music 500) version to the Nucleus version but no mention was made of an automated way.

So that wasn't what I was doing - I was running the BCE version of ample against the B-Em Music 5000 emulator on the basis that the hardware should probably work with both.  It would be interesting to see if the badly behaving BCE tunes works correctly with the compatibility module and Ample Nucleus.  I'll try it out later.

hoglet67:
I just transferred the Music 500 tape to disk and was able to test this on my BeebFPGA with the Music 5000 included.

"Branden" and Syncron" both played correctly.

So it looks like we have at least one more bug to track down in the C emulation.

I'll have a think....It has to be something quite subtle. I wonder if wrap around from Channel 15 back to Channel 0 is correctly implemented?

hoglet67:
I've fixed the bug that was preventing BRANDEN from playing correctly.
https://github.com/hoglet67/b-em/commit/69f229145a2c48203bff2b2081bff8dc71826452

I don't really understand the feature BRANDEN is using that relies on this behaviour. I also tried to list BRANDEN but there is not much there. Do you know if this is possible, or is this somehow protected?

SteveFosdick:
AMPLE words are converted to an intermediate code so all LIST shows is the word currently being edited.  Most tunes seem to ship with this as mostly comments and a call to a word "play".  You can decompile the words back to editable form with EDIT, as per the screendump below:
screenshot from 2017-03-06 13-24-45

hoglet67:
The previous fix didn't completely fix the SYNCRON track.

I ran through some examples in the User Guide, and the Wha sound, which uses Synchronization modulation, was broken.

This fixes that:
https://github.com/hoglet67/b-em/commit/ed64b0bacfe950c29777b847449bad9fdb553590

I've created a new pull request (https://github.com/stardot/b-em/pull/14) that contains these two fixes, plus the Music500 system disk.

PR #14, opened 2017-03-06
Db music5000 fixes

hoglet67:
Here are some fixes to some of the more complex forms of Music 5000 modulation that were not properly implemented.

For reference, I've also added the a working Music 500 system disk, which contains the tracks "BRANDEN" and "SYNCRON" that didn't previously play correctly.


ThomasAdam (approved):
Looks good.

Are you using Gerrit someplace, @hoglet67?

hoglet67:
@ThomasAdam -- no I'm not currently using Gerrit for any of my projects.

Do you think it's something we should use?

ThomasAdam:
No, I will plumb in travis-ci when some of the dependencies go through.

I asked about Gerrit because if the Change-Id lines in your commit messages which is what Gerrit uses for tracking commits.

hoglet67:
I did use Gerrit a few years ago on a work project. I guess I must have somehow installed the Change-ID commit hook globally. Actually, it's sometimes quite useful to have an ID that doesn't change when you rebase, and it also lets you keep track of when you have cherry-picked a commit (not that I seem to do that very often).

SteveFosdick:
@ThomasAdam How far are you intending to take CI?  Just that it builds?  if we want to also run a test suite we will need to add a command line option to autoboot a disc image or some such thing and a mechanism for code on the BBC to send pass/fail messages back to the test harness.

ThomasAdam:
So right now -- I was considering CI as nothing more than compilation builds on different platforms (Linux) and compilers (GCC/Clang).  I've already started this work in `ta/clang` -- problem is that we require dependencies in the form of allegro4 which isn't in the repositories sanctioned by Travis-CI due to S{G,U}IDs.

I'm in talks with Travis about this.

PR #15, opened 2017-03-06
VDFS, SCSI and Config Fixes

SteveFosdick:
This branch loads/saves the state of VDFS and SCSI from the config file, guards against some string buffer overflow errors, and a couple of VDFS-specific fixes.

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -10,10 +10,12 @@
 #include "keyboard.h"
 #include "mouse.h"
 #include "ide.h"
+#include "scsi.h"
 #include "sn76489.h"
 #include "sound.h"
 #include "tape.h"
 #include "tube.h"
+#include "vdfs.h"
 #include "video_render.h"
 
 int curmodel;
@@ -74,7 +76,9 @@ void config_load()
         
         fasttape        = get_config_int(NULL, "fasttape",      0);
 
+	scsi_enabled    = get_config_int(NULL, "scsienable", 0);
         ide_enable      = get_config_int(NULL, "ideenable",     0);
+	vdfs_enabled    = get_config_int(NULL, "vdfsenable", 0);
         
         keyas           = get_config_int(NULL, "key_as",        0);
 
@@ -125,7 +129,9 @@ void config_save()
 
         set_config_int(NULL, "fasttape", fasttape);
 
+	set_config_int(NULL, "scsienable", scsi_enabled);
         set_config_int(NULL, "ideenable", ide_enable);
+	set_config_int(NULL, "vdfsenable", vdfs_enabled);
         
         set_config_int(NULL, "key_as", keyas);
         
--- a/src/config.c
+++ b/src/config.c
@@ -27,7 +27,7 @@ void config_load()
         int c;
         char s[256];
         char *p;
-        sprintf(s, "%sb-em.cfg", exedir);
+        snprintf(s, sizeof(s), "%sb-em.cfg", exedir);
         //printf("%s\n",s);
         set_config_file(s);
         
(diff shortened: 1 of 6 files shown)


src/config.c
```diff
@@ -27,7 +27,7 @@ void config_load()
         int c;
         char s[256];
         char *p;
-        sprintf(s, "%sb-em.cfg", exedir);
+        snprintf(s, sizeof(s), "%sb-em.cfg", exedir);
```
ThomasAdam, line 32: Barely, since you're not checking the return value.  Better would be a wrapper around this, but don't bother.  I just want to point out this change is no better than leaving it as it is.

SteveFosdick:
Except as-was if it overflows the buffer it may just do something random.  Now, though there is no error message, the only ill-effect if the buffer is too small is the pathname will be truncated and probably won't point to a file that can be opened but at least it is predictable.

Anyway, I only made the change because I saw it in passing.  The real work is to add the load/save of the vdfs and scsi flags.

SteveFosdick:
And, we could re-write that function but if we go for paths from autotools rather than the same directory the executable is in this bit would be re-written anyway.

hoglet67:
@SteveFosdick -- are SCSI hard disks on B-Em currently working? I have just tried the latest code, and just seem to get "Broken directory".

It's probably user error. I just copied over my working scsiN.dat and scsiN.dsc files from BeebEm, and expected them to work.

It's probably worth writing a few notes on what steps need to be taken.

I'm on Windows at the moment, if that makes a difference, I'm using Master 128 (ADFS 1.50) with the SCSI hard disk option enabled.

hoglet67:
Ah, I see from the Unix error log they are expected to be in a scsi/ directory.
scsi lun 0: unable to open data file scsi/scsi0.dat: No such file or directory
scsi lun 1: unable to open data file scsi/scsi1.dat: No such file or directory
scsi lun 2: unable to open data file scsi/scsi2.dat: No such file or directory
scsi lun 3: unable to open data file scsi/scsi3.dat: No such file or directory

Do we just loose the error log on windows?

hoglet67:
OK, that's working fine now on Windows.

One question: what's the impact of the .dsc file being inconsistent with the .dat file (i.e. declaring the wrong geometry?) Is this only really used by a formatter?

SteveFosdick:
The geometry details are returned by the ModeSense SCSI command.  I don't think ADFS needs them during normal operation, though it may rely on the total number of sectors being correct.  Formatter?  Maybe.  I don't think I created a SCSI hard disc file from with the BBC, there's a C program in the B-Em src directory, hdfmt.c, that will do that from the host including writing the necessary ADFS structures to it.

The SCSI code in B-Em does expect the sector count derived by multiplying the geometry figures to be correct but it will tolerate the .dsc file not being there and, if it is missing, or the total sector count from the .dsc file is zero, it will use the size of the host file instead and write a .dsc file that gives the corresponding number of sectors.

The reason for using the sector count from the .dsc file is so that a "seed" hard disc file can contain just the used sectors at the start and then grow as more sectors get written to, but report the correct size to the guest from the start.

SteveFosdick:
On the question of logging, that message is logged as a warning (bem_warnf).  The implementation in logging.c just writes that to stderr.  I am not sure why stderr is visible on Unix and not on Windows.  If starting a program from a command prompt I would expect stderr to go to the command prompt window.

I know is is more common on Windows to pop up a message box but I wanted to avoid that for anything but errors as there is nothing worse than going to use a program and half a dozen message boxes appear before you can get going.  There is the option, though, of making the logging functions platform specific so they behave differently.

Issue #17, opened 2017-03-08
Warning logging on windows is not visible


hoglet67:
Instead of sending this to stderr, we should probably write this to a file.

(This is what BeebEm does)

SteveFosdick:
Ok, leave that one with me.

SteveFosdick:
Pull request #19 is in to fix this.  Out of the box the new logging setup will write warnings to the b-emlog.txt file.  To have warnings popup as dialog boxes add "logging = 1905" to the b-em.cfg file.  At the moment there is no GUI to configure the logging - we could add this later is there is demand.

SteveFosdick:
Now pull request #21

SteveFosdick:
This is closed by merging the sf/logging branch.

PR #21, opened 2017-03-09
Sf/logging: Fix Issue #17 / More Flexible logging

SteveFosdick:
The mejor piece of this is a re-write of the logging.c module so it can send message of each class (DEBUG, WARNING, ERROR) to each of three destinations (log file, stderr, message box) according to a config variable.  In the process it fixes Issue '17 by enabling warnings to go to the log file as default behaviour.

### DIFF
--- a/src/Makefile.win
+++ b/src/Makefile.win
@@ -4,7 +4,7 @@ CPP  = g++.exe
 CC   = gcc.exe
 WINDRES = windres.exe
 
-CFLAGS = -O3 -DBEM -DVERSION=\"Win32\"
+CFLAGS = -O3 -Wall -DBEM -DVERSION=\"Win32\"
 
 OBJ = \
     6502.o \
--- a/src/b-em.h
+++ b/src/b-em.h
@@ -40,6 +40,8 @@ extern void bem_error(const char *s);
 extern void bem_errorf(const char *fmt, ...) printflike;
 extern void bem_warn(const char *s);
 extern void bem_warnf(const char *fmt, ...) printflike;
+extern void bem_info(const char *s);
+extern void bem_infof(const char *fmt, ...) printflike;
 extern void win_log_msgbox(const char *level, const char *s);
 
 // If debugging is enabled a real pair of functions will be available
--- a/src/b-em.h
+++ b/src/b-em.h
@@ -34,10 +34,13 @@ void setejecttext(int drive, char *fn);
 #define printflike
 #endif
 
-extern void bem_error(char *s);
+extern void log_open(void);
+extern void log_close(void);
+extern void bem_error(const char *s);
 extern void bem_errorf(const char *fmt, ...) printflike;
 extern void bem_warn(const char *s);
 extern void bem_warnf(const char *fmt, ...) printflike;
+extern void win_log_msgbox(const char *level, const char *s);
 
 // If debugging is enabled a real pair of functions will be available
 // to log debug messages.  if debug is disabled we use a static inline
@@ -47,14 +50,10 @@ extern void bem_warnf(const char *fmt, ...) printflike;
 #ifdef _DEBUG
 extern void bem_debug(const char *s);
 extern void bem_debugf(const char *format, ...) printflike;
-extern void debug_open(void);
-extern void debug_close(void);
 #else
 static inline void bem_debug(const char *s) {}
 static inline void bem_debugf(const char *format, ...) printflike;
 static inline void bem_debugf(const char *format, ...) {}
-static inline void debug_open(void) {}
-static inline void debug_close(void) {}
 #endif
 
 extern char exedir[512];
--- a/src/compat_wrappers.c
+++ b/src/compat_wrappers.c
@@ -40,19 +40,6 @@ x_fopen(const char *path, const char *mode)
 #include <errno.h>
 
 int
-asprintf(char **ret, const char *fmt, ...)
-{
-	va_list	ap;
-	int	n;
-
-	va_start(ap, fmt);
-	n = vasprintf(ret, fmt, ap);
-	va_end(ap);
-
-	return (n);
-}
-
-int
 vasprintf(char **ret, const char *fmt, va_list ap)
 {
 	int	 n;
@@ -81,6 +68,20 @@ error:
 	*ret = NULL;
 	return (-1);
 }
+
+int
+asprintf(char **ret, const char *fmt, ...)
+{
+	va_list	ap;
+	int	n;
+
+	va_start(ap, fmt);
+	n = vasprintf(ret, fmt, ap);
+	va_end(ap);
+
+	return (n);
+}
+
 #endif
 
 #ifndef HAVE_STPCPY
--- a/src/disc.c
+++ b/src/disc.c
@@ -63,9 +63,9 @@ void disc_load(int drive, char *fn)
         setejecttext(drive, "");
         if (!fn) return;
         p = get_extension(fn);
-        if (!p) return;
+        if (!p || !*p) return;
         setejecttext(drive, fn);
-        bem_debugf("Loading :%i %s %s\n", drive, fn,p);
+        bem_infof("disc: Loading %i %s %s\n", drive, fn, p);
         while (loaders[c].ext)
         {
                 if (!strcasecmp(p, loaders[c].ext))
--- a/src/fdi2raw.c
+++ b/src/fdi2raw.c
@@ -1311,7 +1311,7 @@ static int handle_sectors_described_track (FDI *fdi)
 {
 	uae_u8 *start_src = fdi->track_src;
 #ifdef _DEBUG
-	int oldout;
+	int oldout = 0;
 #endif
 	fdi->encoding_type = *fdi->track_src++;
 	fdi->index_offset = get_u32(fdi->track_src);
@@ -1328,7 +1328,7 @@ static int handle_sectors_described_track (FDI *fdi)
 		oldout = fdi->out;
 #endif
 		if (fdi->out < 0 || fdi->err) {
-			bem_debugf("\nin %ld bytes, out %d bits\n", fdi->track_src - fdi->track_src_buffer, fdi->out);
+		        bem_debugf("\nin %ld bytes, out %d bits\n", (long)(fdi->track_src - fdi->track_src_buffer), fdi->out);
 			return -1;
 		}
 		if (fdi->track_src - fdi->track_src_buffer >= fdi->track_src_len) {
--- a/src/linux.c
+++ b/src/linux.c
@@ -50,11 +50,6 @@ void updatewindow()
 {
 }
 
-void bem_error(char *s)
-{
-        allegro_message(s);
-}
-
 void setquit()
 {
         quited=1;
@@ -67,7 +62,7 @@ int main(int argc, char *argv[])
 
         if (allegro_init())
         {
-                printf("Failed to initialise Allegro!");
+	        fputs("Failed to initialise Allegro!\n", stderr);
                 exit(-1);
         }
 
--- a/src/logging.c
+++ b/src/logging.c
@@ -2,81 +2,153 @@
 
 #include "b-em.h"
 
+#include <allegro.h>
 #include <errno.h>
 #include <stdarg.h>
+#include <time.h>
+
+#define LOG_DEST_FILE   0x01
+#define LOG_DEST_STDERR 0x02
+#define LOG_DEST_MSGBOX 0x04
+
+#define LOG_DEBUG_MASK  0x000f
+#define LOG_DEBUG_SHIFT 0
+#define LOG_WARN_MASK   0x00f0
+#define LOG_WARN_SHIFT  4
+#define LOG_ERROR_MASK  0x0f00
+#define LOG_ERROR_SHIFT 8
+
+static unsigned log_options =
+    (LOG_DEST_FILE << LOG_DEBUG_SHIFT) |
+    ((LOG_DEST_FILE|LOG_DEST_STDERR) << LOG_WARN_SHIFT) |
+    ((LOG_DEST_FILE|LOG_DEST_MSGBOX) << LOG_ERROR_SHIFT);
+
+static const char log_fn[] = "b-emlog.txt";
+FILE *log_fp;
+
+static char   tmstr[20];
+static time_t last = 0;
+
+static void log_common(unsigned dest, const char *level, const char *msg, size_t len)
+{
+    time_t now;
+
+    while (msg[len-1] == '\n')
+	len--;
+    if ((dest & LOG_DEST_FILE) && log_fp) {
+	time(&now);
+	if (now != last)
+	{
+	    strftime(tmstr, sizeof(tmstr), "%d/%m/%Y %H:%M:%S", localtime(&now));
+	    last = now;
+	}
+	fprintf(log_fp, "%s %s ", tmstr, level); 
+	fwrite(msg, len, 1, log_fp);
+	putc('\n', log_fp);
+	fflush(log_fp);
+    }
+    if (dest & LOG_DEST_STDERR) {
+	fwrite(msg, len, 1, stderr);
+	putc('\n', stderr);
+    }
+    if (dest & LOG_DEST_MSGBOX) {
+#ifdef WIN32
+	win_log_msgbox(level, msg);
+#else
+	alert(level, msg, "", "&OK", NULL, 'a', 0);
+#endif
+    }
+}
+
+static void log_plain(unsigned mask, unsigned shift, const char *level, const char *msg)
+{
+    unsigned opt = log_options & mask;
+
+    if (opt)
+	log_common(opt >> shift, level, msg, strlen(msg));
+}
+
+static const char msg_malloc[] = "log_format: out of space - following message truncated";
+
+static void log_format(unsigned mask, unsigned shift, const char *level, const char *fmt, va_list ap)
+{
+    unsigned opt, dest;
+    char   abuf[200], *mbuf;
+    size_t len;
+
+    if ((opt = log_options & mask))
+    {
+	dest = opt >> shift;
+	len = vsnprintf(abuf, sizeof abuf, fmt, ap);
+	if (len <= sizeof abuf)
+	    log_common(dest, level, abuf, len);
+	else if ((mbuf = malloc(len + 1))) {
+	    vsnprintf(mbuf, len, fmt, ap);
+	    log_common(dest, level, mbuf, len);
+	    free(mbuf);
+	} else {
+	    log_common(dest, level, msg_malloc, sizeof msg_malloc);
+	    log_common(dest, level, abuf, len);
+	}
+    }
+}
 
 #ifdef _DEBUG
-static const char debug_fn[] = "b-emlog.txt";
-FILE *debug_fp;
 
 void bem_debug(const char *s)
 {
-        if (debug_fp)
-        {
-                fputs(s, debug_fp);
-                fflush(debug_fp);
-        }
+    log_plain(LOG_DEBUG_MASK, LOG_DEBUG_SHIFT, "DEBUG", s);
 }
 
 void bem_debugf(const char *fmt, ...)
 {
-        va_list ap;
+    va_list ap;
 
-        if (debug_fp)
-        {
-                va_start(ap, fmt);
-                vfprintf(debug_fp, fmt, ap);
-                va_end(ap);
-                fflush(debug_fp);
-        }
+    va_start(ap, fmt);
+    log_format(LOG_DEBUG_MASK, LOG_DEBUG_SHIFT, "DEBUG", fmt, ap);
+    va_end(ap);
 }
 
 #endif
 
-void bem_errorf(const char *fmt, ...)
-{
-        char buf[256];
-        va_list ap;
-
-        va_start(ap, fmt);
-        vsnprintf(buf, sizeof buf, fmt, ap);
-        va_end(ap);
-
-        bem_error(buf);
-        bem_debug(buf);
-}
-
 void bem_warn(const char *s)
 {
-        fputs(s, stderr);
-        fputc('\n', stderr);
+    log_plain(LOG_WARN_MASK, LOG_WARN_SHIFT, "WARNING", s);
 }
 
 void bem_warnf(const char *fmt, ...)
 {
-        char buf[256];
-        va_list ap;
+    va_list ap;
 
-        va_start(ap, fmt);
-        vsnprintf(buf, sizeof buf, fmt, ap);
-        va_end(ap);
-
-        bem_warn(buf);
-        bem_debug(buf);
+    va_start(ap, fmt);
+    log_format(LOG_WARN_MASK, LOG_WARN_SHIFT, "WARNING", fmt, ap);
+    va_end(ap);
 }
 
-#ifdef _DEBUG
-
-void debug_open()
+void bem_error(const char *s)
 {
-        if ((debug_fp = fopen(debug_fn, "wt")) == NULL)
-                bem_warnf("unable to open debug log %s: %s", debug_fn, strerror(errno));
+    log_plain(LOG_ERROR_MASK, LOG_ERROR_SHIFT, "ERROR", s);
 }
 
-void debug_close(void)
+void bem_errorf(const char *fmt, ...)
 {
-        if (debug_fp)
-                fclose(debug_fp);
+    va_list ap;
+
+    va_start(ap, fmt);
+    log_format(LOG_ERROR_MASK, LOG_ERROR_SHIFT, "ERROR", fmt, ap);
+    va_end(ap);
 }
 
-#endif
+void log_open(void)
+{
+    log_options = get_config_int(NULL, "logging", log_options);
+    if ((log_fp = fopen(log_fn, "at")) == NULL)
+	bem_warnf("log_open: unable to open log %s: %s", log_fn, strerror(errno));
+    bem_debugf("log_open: log options=%d", log_options);
+}
+
+void log_close(void)
+{
+    if (log_fp)
+	fclose(log_fp);
+}
--- a/src/logging.c
+++ b/src/logging.c
@@ -13,13 +13,16 @@
 
 #define LOG_DEBUG_MASK  0x000f
 #define LOG_DEBUG_SHIFT 0
-#define LOG_WARN_MASK   0x00f0
-#define LOG_WARN_SHIFT  4
-#define LOG_ERROR_MASK  0x0f00
-#define LOG_ERROR_SHIFT 8
+#define LOG_INFO_MASK   0x00f0
+#define LOG_INFO_SHIFT  4
+#define LOG_WARN_MASK   0x0f00
+#define LOG_WARN_SHIFT  8
+#define LOG_ERROR_MASK  0xf000
+#define LOG_ERROR_SHIFT 12
 
 static unsigned log_options =
     (LOG_DEST_FILE << LOG_DEBUG_SHIFT) |
+    (LOG_DEST_FILE << LOG_INFO_SHIFT)  |
     ((LOG_DEST_FILE|LOG_DEST_STDERR) << LOG_WARN_SHIFT) |
     ((LOG_DEST_FILE|LOG_DEST_MSGBOX) << LOG_ERROR_SHIFT);
 
@@ -110,6 +113,20 @@ void bem_debugf(const char *fmt, ...)
 }
 
 #endif
+
+void bem_info(const char *s)
+{
+    log_plain(LOG_INFO_MASK, LOG_INFO_SHIFT, "INFO", s);
+}
+
+void bem_infof(const char *fmt, ...)
+{
+    va_list ap;
+
+    va_start(ap, fmt);
+    log_format(LOG_INFO_MASK, LOG_INFO_SHIFT, "INFO", fmt, ap);
+    va_end(ap);
+}
 
 void bem_warn(const char *s)
 {
--- a/src/main.c
+++ b/src/main.c
@@ -111,7 +111,7 @@ void main_init(int argc, char *argv[])
         int c;
         int tapenext = 0, discnext = 0;
 
-        debug_open();
+        log_open();
 
         startblit();
         
@@ -120,7 +120,6 @@ void main_init(int argc, char *argv[])
 	vid_fskipmax = 1;
         
         al_init_main(argc, argv);
-        
         
         append_filename(t, exedir, "roms\\tube\\ReCo6502ROM_816", 511);
         if (!file_exists(t,FA_ALL,NULL) && selecttube == 4) selecttube = -1;
@@ -393,7 +392,7 @@ void main_close()
         
         al_close();
         video_close();
-        debug_close();
+        log_close();
 }
 
 void changetimerspeed(int i)
--- a/src/main.c
+++ b/src/main.c
@@ -112,6 +112,7 @@ void main_init(int argc, char *argv[])
         int tapenext = 0, discnext = 0;
 
         log_open();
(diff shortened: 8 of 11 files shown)


ThomasAdam:
Hi,

Thanks, @SteveFosdick.  Whilst I appreciate the need for logging, I'm starting to think this is rather overboard.  I would suggest the following:

* Turning on logging implies just that -- that information is sent to a text file.  There should be just one interface for that -- say:  `log_msg` or somesuch.  That can be variadic, and we can append/prepend information to that line if necessary.  If logging isn't enable that function call does nothing.

* A fatal error (that might require a GUI popup) can make use of the current `bem_*` interface as before.

I am really not a fan of having all these levels, and right now there's a tonne of complexity in these changes which I'm not sure we require.

SteveFosdick:
But that fails to address the case where the person using it should be aware of something but it is reasonable to continue.  In windows stderr and stdout go nowhere.  Applying an oversimplistic approach to opening files has already meant that now b-em crashes when it can;'t open a disc image whereas previously it would have continued as if no disc was in the drive.

fordp2002:
As b-em is under active development powerful tracing capabilities are surely welcome. As Steve has put effort in to improving b-em what is the harm in merging it as long as Steve is willing to maintain this new feature until it is at least stable?

ThomasAdam:
@fordp2002 -- I'm not disputing the need for logging.  What I'm reviewing is how that logging is architected.   You might see this as a barrier to entry; I see this as a way of discussing the implications of the proposed changes, which is all I am interested in doing here.  I won't blindly merge changes just because there's been effort put in -- nor will I seemingly "block" a piece of code review from going through if I disagree with it.  I made it clear --- very clear --- when I took up the reins here that things would be transparent, and this it the entire purpose for code review.   To have that conversation.  I hope you can see the benefit to it.  As for merging prematurely just because work has been done... no.  I can't have that.   Now is the appropriate time to have that conversation.

@SteveFosdick -- I understand your point about stderr/stdout seemingly not going anywhere under Windows.  I'd be keen to understand why -- but really, my concerns are over the complex logging levels.  In my mind, if someone enables logging, they either get everything, or nothing -- and hence, all the states of WARN/INFO/ERR can go away.

As for what should happen on Windows --- then I can well imagine a need for a dedicated file.  That's probably OK.  Again, as with the UNIX side of things -- that can entail a general-purpose log file and where necessary a separate mechanism for generating GUI windows for errors which are best represented in that way.  But I still think, regardless, there needs to be only two interfaces to logging:

* One which is writing to a file (stderr on UNIX/something else on Windows):  Maybe `log_debug(...)`

* One which generates a GUI box with said message.

If this isn't clear, I'm more than happy to code up what I'm referring to.

Kindly,
Thomas

fordp2002:
On the many millions of lines code I work on for my day job the diagnosis logging is way more sophisticated than you describe. You can filter on severity, module and sub module and also direct the output to the screen or files and view it locally or remote​ly. We regularly both need and use this functionality. This is akin to drill down in the accounting software world. I am not saying we need that for B-Em, but just file or pop-up may well be too little.

ThomasAdam:
@fordp2002 -- and I'm saying that such an interface is left to the programmer to decide what the appropriate logging level should be.  Even more so is the point that now, that has to be exposed to the user to decide which severity levels they end up logging.

I do not believe at all that b-em is sufficiently advanced such that any different logging levels need to be made separate.  Again, I believe you log everything via the standard API for logging if that's enabled, or not at all, and hence, there's still no loss of data.

As for ancillary things such as corefiles, etc., they can still be analysed in the usual way, but no amount of different logging levels will cure that -- and b-em is not such an application where remote logging, etc., is even necessary.  That said, I doubt there's anything stopping you shunting these outputs to the Windows Event Log, if you so wanted.

fordp2002:
Indeed I am sure B-Em does not need all that complexity. My work stuff can have many thousands of processors spread over buildings or whole sites. Lucky for me most of the nodes are pretty simple.

SteveFosdick:
Message logging of this is type has been a feature of most things I have worked on over the last 25+ years and much more complicated versions have been implemented in frameworks such as log4j and its derivatives.

My thinking was that when writing one of the non-GUI modules in the emulator one should not be deciding whether a particular message pops up in a dialog box, gets written to a file, to stderr or even discarded.  Instead you should be labelling that message with an importance and handing it off to something that will then make that decision and do the appropriate thing.

Once that decision-making is localised to a single module it could be hard-coded, and that's essentially how it was before this change, or it could be configurable.  There is very little code specifically concerned with making it configurable.  That means, for example, that B-em could be run as part of an automated test harness when it is important that execution does not pause awaiting a user to dismiss a dialog box, and interactively when a dialog box might be helpful without having to use conditional compilation or add any code.

It is also worth noting that sensible defaults are built-in so no-one has to configure customised logging if he doesn't want to and the default should be suitable for Linux and Windows.  As you probably gather I don't subscribe to the philosophy that configuration is evil.  I prefer to have things work sensibly "out of the box" but configurable if people want to.

hoglet67:
I also feel it's important to have a fairly flexible logging system.

I disagree that this is adding complexity. I would argue the opposite. If we get this right, then modules using this become simpler because there is a uniform pattern adopted everywhere, and people are not afraid to make good use of debug logging, and then leave that in place.

Here's what I think is important:
- Multiple levels (DEBUG, INFO, WARNING, ERROR, FATAL) indicating importance
- The ability to disable DEBUG at compile time, so there is minimal performance penalty (this is important!)
- Include two executables in a standard distribution: the normal one with DEBUG disabled and a debugging one with DEBUG enabled
- Target selection via configuration (tweaking this from the UI can come later)
- A target that is a log file (TBD whether this is reset or rotated on startup)
- A second target that is a log window (Atomulator does this currently) 
- A sensible default configuration that works on Windows and Linux

I'm less sure about the utility of pop-up dialog boxes for log messages. When things go wrong, you tend to get a lot of logging and this could easily overwhelm the user.

I share the view that a flexible logging framework is important to include in B-Em, especially as we anticipate lots of development happening by lots of people. In some of the CPU emulations in PiTubeDirect we could have benefited from such a framework. What we had was very adhoc, and in one case it looked like the code included lots of checks and warning messages, but these were actually all being disabled at compile time due to some complicated #ifdefs.

Also, having a common logging API like this will help portability of sub-systems between B-Em and other projects, like BeebEm and PiTubeDirect.

This is actually pretty similar to what Steve has prototyped so far, I would suggest we continue to develop this, with a view to merging as soon as the APIs are stable.


Although not really logging, I would encourage us to also consider a CPU execution tracing facility. This is something I might start working on in PiTubeDirect. The goal would be to be able to capture trace (with disassembly) of executing code, from any of the CPU emulations. This also needs to be very configurable, but you generally want to be able to define a trigger "event" e.g. PC=&802D, and decide whether you want this event to be at the start, middle, or end of a large capture window. The capture buffer would be in memory, so there is minimal impact on execution speed. Once full, it is dumped to a file.

fordp2002:
Thanks Dave,
The number one thing I would like to add to B-Em is a 32016 Instruction Trace/ Disassembler/ Control Console. I must admit I would like to do it for the sheer hell of it and as a learning process for myself. I guess others would love a similar system for the 6502. I am happy to collaborate.

I will not proceed however if I feel my contributions are not going to be accepted in to the star dot version.

I contributed hundreds of hours in to another Git Hub project only to find that non of my changes were accepted upstream. I stopped working on that project as soon as I realised that the only way people were going to get my features was from my Fork and that there was little chance of my changes ever being combined with the efforts of others.

SteveFosdick:
Dave on the subject of CPU tracing I did implement that for the I/O 6502 though checking that doesn't seem to be in the stardot versiob.  That might be my fault for not submitting it.  It logs addresses, mnemonics and register contents to a text file.  It can't be set to automatically start/stop at particular addresses, so that a particular region is traced whenever it is executed, but something similar can be done manually by setting a breakpoint at the start, at which point you can issue a debugger command to enable tracing, set a breakpoint for the stop tracing point, continue, and then disable tracing.

I can put in in a branch and put in a pull request for it if people think it would be useful but I'd rather not have more than one pull request from me outstanding at a time because I seem to get into a muddle with git if I do.  If we then want to enhance that further we can - my preference is for things to be merged as soon as usable rather that wait for every bell and whistle to be there.

SteveFosdick:
Actually, to correct myself, the CPU tracing is there in the stardot version - I was checking the wrong source file - though I haven't updated the help.  To use start b-em with -debug then from the debugger prompt `trace filename` to turn on, `trace` without the filename to turn off.

ThomasAdam:
So on top of this pull-request, I've come along and butchered it to show my intent.  See:

https://github.com/stardot/b-em/commit/4e15018a491e3aef9c2476f6758a9f088d1c2cfe

I suppose the point here is that:

* Different log-levels are retained -- but it is reduced down to one API call; the decision is still left up to the programmer to decide which log level is best for whatever piece of information they're trying to convey, but it happens via `bem_log()` now.
* Log-levels are cumulative (if you set `logging = 1` in `b_em.cfg` then you'll see both `INFO` and `WARN` messages, for instance.
* No explicit logging to `stderr` on UNIX.  If you want that, `tail` the log file instead; this reduces the complexity of having `bem_*f()` counterpart functions.
* I've broken the MSGBOX stuff for Windows -- sorry.
* I still am unsure whether a loglevel of `LOG_ERROR` should even exist -- by that point it's fatal and we'll be erroring out anyway.  But it's kept unchanged for now.

How does this look?  Mainly, I'm trying to:

* Reduce the number of logging APIs to reduce down to a single public instance;
* No need for bit-shifting to work out different logging scenarios (just use an `enum` position to mark the level;

What do people think?

SteveFosdick:
My first point is a general one on the calling of exit().  There are very few situation is which it is necessary to exit and as far as I can see those are limited to failure to initialise Allegro, failure to initialise one of the modules that emulate the hardware or failure to load the OS ROM.  For every other kind of failure it would be better for the emulator to continue, particularly as quitting unexpectedly could cause someone using the emulator to lose work.

As such I don't believe that failure to open the log file should cause an exit, rather stderr should be used as the stdio stream to which messages are directed instead, the first of which would be a warning that the proper log file could not be opened.  Failure to allocate memory for a temporary buffer in which to format the log message should not be a cause for an exit either.  Yes, it may mean the program is nearly out of memory but as there are few other places in which memory is dynamically allocated that does not mean it is about to crash and even if it was, a message, maybe a static one that doesn't need a malloc call, to the log file to indicate that an out of memory condition was the problem would be very helpful.

So when I wrote:

```
	len = vsnprintf(abuf, sizeof abuf, fmt, ap);
	if (len <= sizeof abuf)
	    log_common(dest, level, abuf, len);
	else if ((mbuf = malloc(len + 1))) {
	    vsnprintf(mbuf, len, fmt, ap);
	    log_common(dest, level, mbuf, len);
	    free(mbuf);
	} else {
	    log_common(dest, level, msg_malloc, sizeof msg_malloc);
	    log_common(dest, level, abuf, len);
	}
```

it was not that I didn't know of vasprintf but that I chose to write something that would both be faster, in that for the majority of messages it would not need either to call malloc or to process the printf formatting twice, one to get the length and once to actually do it, and would be more robust because in the occasional case of a long message that did require a malloc, if the malloc failed it would print a static message about the malloc failure and then the truncated version of the original message to the log.

Why do we need 2048 characters to format a date/time into?

If going for levels rather than masks with the lowest level being the highest priority shouldn't the order be ERROR,WARNING,INFO,DEBUG?

When reading the desired log level from the config file, if the result is out of range it would be better to clamp the value to in-range, so to ERROR if negative and to DEBUG if greater numerically than DEBUG?

I am not particularly bothered by the demise of the non-printf-like functions as the majority of messages do seem to use formatting.  As far as style goes, I don't have a preference one way or the other over separately named functions (bem_debug, bem_error etc) or your scheme with an extra argument though your scheme doesn't make it possible to use conditional compilation to remove the debug calls alltogether.

x_fopen no longer needs to call asprintf itself - it can use the formatting via bem_log.

On the subject of x_fopen there is a bug in ssd.c (and probably in dsd.c too) where the code tries to open the image file read-write first and then falls back to opening it read-only if that fails.  Substituting x_fopen for fopen has broken that so now if an image file is read-only at OS level B-em will crash.  Likewise crashing out when unable to re-load the SSD from a previous session is a step backwards in behaviour.

Regards,
Steve.

ThomasAdam:
OK.  Thanks.

* On the use of `exit` -- that's a different problem, although I will argue quite strongly that there are many instances where failures in things like `malloc` really is fatal.  To continue from that could be very bad.  Just because one instance might not need to malloc, doesn't mean it should be ignored as such; something else could come along and then land things in more trouble,  As for the other instances you're referring to, pass.  I'll have to look at some point.
* I'll change the log file opening check to be less obtrusive.
* We don't need 2048 characters for a date/time -- I'll reduce it, but it's just a number.
* As for `x_fopen()` checks in ssd.c -- yuck.  This has nothing to do with `x_fopen()`, if you want two different modes, use `access()`.  I'll change this.

Kindly,
Thomas

SteveFosdick:
On the question of malloc and exit, in general, programs may indeed crash shortly after the first malloc failure but b-em is a program that makes very little use of dynamic memory so that doesn't apply here.

On using access in ssd/dsd, access is not the answer. On linux it checks the permissions using the real uid/gid whereas open, which fopen will ultimately call, uses the effective uid/gid so one does not predict the outcome of the other. I don't believe the original code before x_fopen was introduced could be improved. It could use the current setting of the drive write-protect flag to open the file in the appropriate mode but as we allow the user to dynamically change that flag, that approach would force reopening the file on a change from write protect to not. The original code also had the correct behavior - if neither attempt to open the file worked carry on and emulate an empty drive. Now it quits and looses what the user was doing instead.

hoglet67:
> I am not particularly bothered by the demise of the non-printf-like functions as the majority of messages do seem to use formatting. As far as style goes, I don't have a preference one way or the other over separately named functions (bem_debug, bem_error etc) or your scheme with an extra argument though your scheme doesn't make it possible to use conditional compilation to remove the debug calls alltogether.

This last part is very important to me. I want to share code between PiTubeDirect and this project. I need to ensure that debug logging code is not included in a normal build, and there is no performance impact of including it in the code. It's only possible to do this if each log level has a separately named function. that can be removed using conditional compilation.

Stylistically, I also have a strong preference for 5 separately named functions with prinf() type syntax.

There is value in WARNING, ERROR and FATAL:
- WARNING: something has happened that might be of interest to the developers, but the user probably doesn't care
- ERROR: something serious has happened (e.g. an expected file could not be opened), but we'll continue anyway. 
- FATAL: something very serious has happened, and we need to exit

FATAL should always be followed by exit()

fordp2002:
hoglet wrote: "Stylistically, I also have a strong preference for 5 separately named functions with prinf() type syntax."

An alternative that I use is say 5 separate Varidatic macros that call the printf like function.

In my embedded code I have many macros that all equate to printf. These serve two purposes.
1) They highlight the "source" of the message
2) They allow the printf to be cleanly removed for a release build.

e.g In my serial header:

#ifdef DEBUG
#define SERIAL_DEBUG printf
#else 
#define SERIAL_DEBUG(...)
#endif

hoglet67:
Thomas,

There is a real danger here that work on B-Em will stall if we can't get this issue agreed. And I'm also concerned this is all very off-putting to new contributors.

I think first and foremost we need to pin down the API. The initial implementation then just needs to be good enough, and we can improve it over time. My view is that we should start with just this:
```
void log_debug(const char *fmt, ...);
void log_info(const char *fmt, ...);
void log_warn(const char *fmt, ...);
void log_error(const char *fmt, ...);
void log_fatal(const char *fmt, ...);
```
Note, I've changed the suffix from bem_ to log_ as I would like to use this in PiTubeDirect, so we can more easily share code. It just needs to be something generic.

The separate methods allow us to eliminate debug logging using an implementation like this:
```
static inline log_debug(const char *fmt, ...) {};
```
This form has the advantage (I think) that it won't generate variable-not-used warnings, yet the compiler should be able to optimise away all the overhead, including preparing all the parameters.

It might also be useful to have:
```
int log_is_debug_enabled();
```
so more complex blocks of debug could be switched out. I'm very open to alternative suggestions here that achieve the same effect.

Dave

ThomasAdam:
@hoglet67 -- It's not stalled though, we're talking about it which is important.  What I don't want to do is shove something through and then have to rework it.  The whole point of having this conversation now is so that we iron out those things.  Where you see time taken away from usability and features, I see valuable time taken to ensure this is correct -- and having that conversation so far has been really interesting indeed.  This does not prevent anyone from submitting more pull-requests.  If that's an issue for any contributor due to logistics, then again, I'm more than happy to help people manage that.  You'll note that I am quite proactive in merging most people's pull-requests and giving feedback where I think it appropriate and necessary.

As for having parity between b-em and PiTubeDirect, I really do not want to have either project dictate any kind of API or way of working, because either they're separate projects, or they're not.  Changes in b-em should not prohibit PiTubeDirect from working or making things easier.  If that's not the case, then there's a different problem to look at.

As for debug -- I am not a fan of `#ifdef` code all over the place for it.  I would much rather see DEBUG be just another logging level, and all of the `#ifdef` code go away.  I suspect it's not as simple as that; I haven't looked, but it might be.  That would then obviate the need for any crazy separate handling.

As for the API for logging, separate functions is something I'm really keen to avoid, as previously stated.

hoglet67:
> As for having parity between b-em and PiTubeDirect, I really do not want to have either project dictate any kind of API or way of working, because either they're separate projects, or they're not. Changes in b-em should not prohibit PiTubeDirect from working or making things easier.

They are separate projects, but there is a strong desire to make sharing code easier. Agreeing common APIs for logging is part of that. What doesn't make sense about this? Why make life unnecessarily difficult?

> As for the API for logging, separate functions is something I'm really keen to avoid, as previously stated.

Yes, you have asserted this several times, but I don't think you have explained why (sorry if I have missed this).

Can you explain with your single method how you eliminate the overhead of debug logging in a  "production" build?

Maintaining good performance is an issue, certainly with PiTubeDirect, and also with B-Em. I personally run B-Em on a 7 year old laptop, and at times it struggles. I don't think we can assume everyone has sufficiently fast hardware that we can ignore the performance implications of decisions we make.

fordp2002:
I am not a big fan of #ifdef's all over the place in the implementation files either. It is however useful to have macro's in the header files to manage the different build options. As for the relationship between B-Em and Pi Direct I feel you have characterised that wrong. We are talking about sharing the code for a Copro(s) between the two projects. In this case the project is really a CoPro and B-Em and Pi Direct are just environments where these CoPros's can operate. Remember in reality at Acorn these were separate products.

ThomasAdam:
> They are separate projects, but there is a strong desire to make sharing code easier. Agreeing common APIs for logging is part of that. What doesn't make sense about this? Why make life unnecessarily difficult?

It's not "unnecessarily difficult" -- I'd like to try and move away from those negative connotations.  I'm not disagreeing common APIs, but I am curious how if there is some sort of parity with b-em, how you're also reflecting that in PiTubeDirect.  Either way, I'm uncomfortable with what I see as internal things for b-em (logging) having an impact on PiTubeDirect -- and if there comes a time when either project somehow impacts one another, or causes headaches, or forces decisions to appease either one in some way that's beyond some sort of API, then there's no point at all in calling them separate, in my opinion.

> Can you explain with your single method how you eliminate the overhead of debug logging in a "production" build?

No, because logging is logging -- at some point, depending on the log-level and the amount of information written out, there may well be overhead.  I'm not clear on what you're asking.

> Maintaining good performance is an issue, certainly with PiTubeDirect, and also with B-Em. I personally run B-Em on a 7 year old laptop, and at times it struggles. I don't think we can assume everyone has sufficiently fast hardware that we can ignore the performance implications of decisions we make.

Absolutely -- neither have I had any issues sent in which say things are slow for any reason, nor do I see any performance implications from the code being suggested here, and I too have an old laptop and the changes I'm proposing and have made, haven't made it any slower that I've seen.  As for profiling that, well... until that happens, it's at most a "best guess".

hoglet67:
> I am curious how if there is some sort of parity with b-em, how you're also reflecting that in PiTubeDirect. Either way, I'm uncomfortable with what I see as internal things for b-em (logging) having an impact on PiTubeDirect.

Simon put it quite succinctly:
> In this case the project is really a CoPro and B-Em and Pi Direct are just environments where these CoPros's can operate.

In PiTubeDirect we have emulations of several Co Processors. I want to get to a place where the code for those Co Processors can be shared between B-Em and PiTubeDirect. This mighty eventually mean these become seperate git submodules, that both parent projects include. But for now, for ease of development, I'm planning on keeping them in PiTubeDirect and periodically updating a copy in B-Em.

As far as the Co Pro is concerned, both B-Em and PiTubeDirect are environments that host one or more Co Pros, and through well defined APIs make services like logging available to the Co Pro. The environment may (will) have different implementations of those API. But unless we make the APIs the same, it won't be possible to use the same Co Pro code in both places.

I haven't fully thought this out yet, but I think there are four core APIs:
- logging (which we are discussing where)
- tube register access (tube_parasite_read and tube_parasite_write)
- debugging (which Steve has made an excellent proposal for [here](http://www.stardot.org.uk/forums/viewtopic.php?f=4&t=12703))
- and lifecycle (i.e. create, execute, destroy, probably some others)

> No, because logging is logging -- at some point, depending on the log-level and the amount of information written out, there may well be overhead. I'm not clear on what you're asking.

All I want is for the logging API support an implementation where debug logging that is disabled (e.g. in a production build) has zero performance overhead, because it is optimised out at compile time.

A single common method for all log levels does not (and cannot) achieve this.

SteveFosdick:
Thomas, There is no doubt in my mind that implementing a CPU emulation is a significant undertaking and it strikes me as less than wise to make choices that make it harder than it needs to be to share CPU emulations between B-Em and PiTubeDirect.

None of this means that PiTubeDirect has to have exactly the same logging _implementation_ as B-em because, for all I know, there may be no files to write to and output may be to a serial port, but if the _API_ for logging is the same that makes sharing code easier.

Regarding performance, I wouldn't spend many hours optimising every last nanosecond out of code without a confirmed performance issue either, but that doesn't mean that when I and the other two people who have been contributing to this discussion can plainly see that one version of an API has limitations that will affect performance which could not be removed without an API change, and the other version of the API better lends itself to good performance, I shouldn't consciously and deliberate choose the one that lends itself to better performance.  Given the amount of effort that is going into this discussion we really don't want to be back here again in the future.

So on the question of overhead, with separate functions and a very simple bit of conditional compilation limited to just the header file that declares this API, when compiled with debugging disabled the log_debug function can be replaced by an empty static function.  With optimisation enabled a modern compiler will be clever enough not to generate code for the empty body, for the call to that function, and if it is a decent one, any function calls involved in providing arguments to that function (strerror, for example).

By comparison if it is the implementation of log_msg that differs depending on whether debug is enabled, or that decision is only ever made at runtime, the compiler is forced to generate a call to log_msg wherever it occurs, even for LOG_DEBUG messages, and that also means generating any code required to provide the arguments to that call.  Yes, the overhead of actually writing to the file is avoided, as it would be with either approach, but the overhead of making the function call is not avoided.

fordp2002:
This debugging scenario is very common in the embedded world. You hook extra kit to your embedded product to display trace information. The production product (say a toaster) does not have that kit. The debug version of the code is much bigger with printf functionality and all the trace generation so on the debug version you use a bigger more expensive chip and maybe run it faster so the debug overhead does not burn your toast!

You then build the release version and every single bit of overhead is removed by slicing it all out AT BUILD TIME. You then have a cheap as chips products and lots of lovely brown toast ;)

ThomasAdam:
Blimey.

OK, so from what I'm hearing (and please do correct me if I'm wrong -- it's useful discussion):

* A single interface to logging (that is, `bem_log` or some such) is OK, providing the special case of `LOG_DEBUG` can be optimised away at compile time;
* There should be no performance impact with logging enabled for any log-level, as this could somehow be running on embedded/old hardware which might suddenly start to fall apart at the edges due to various limitations;
* There should be some consistency between `b-em` and `PiTubeDirect` in terms of shared code.

Some numbers then:

Currently, all calls we have for `DEBUG`:

```
% git grep -P '^\s*bem_log\(LOG_DEBUG' | wc -l
     148
```

For all other calls:

```
 % git grep -P '^\s*bem_log\(LOG_(?:WARN|INFO|ERROR)' | wc -l
     29
```

Currently, if you were to enable `LOG_DEBUG`, you'd therefore have at worse-case:  `177` log calls.  Given some of those will be dependent on configuration, it's hard to tell how many of that you'd get at once.

Masking out `DEBUG` as something special means that:

- Either you have a separate `bem_log_debug` call and compile-time that out.

Or:

- Break all the functions out to their separate counterparts (as was sort of done previously), and suck up the cost of having to add/remove additional logging levels if there are any.

I'm disinclined with the former, if only because it requires boiler-plate wrappers.  As to the latter, that's probably the way to go -- but note that I don't like that either.  If there's such a song-and-dance about DEBUG mode being a performance hit now, will anyone even enable it for fear of seven-plus year old hardware suffering?  I'm being tongue-in-cheek with that statement.  Without any concrete proof that not doing any of these two suggestions incurs a performance hit (and I'd have to see numbers for this), I am unsure whether sacrificing DEBUG as some special case is worth the overhead in allowing that to dictate an API, or worse yet, allowing it to be revered as something which is intended to deluge information -- whilst meaning it's only-ever a compile-time because somehow, it might cripple people's machines.

Mind you, I'm still struggling to even care about the granularity of the log-levels which are being proposed now.  I, personally, couldn't care less for the difference between `INFO` and `WARN` as a user; as a programmer though, I have to care **a lot** to ensure I give a fair chance to the poor soul bothering to turn logging on.  But that's moot -- there's clearly a drive for having such granularity, so I see no need to consider that changing.

So... back to my original point.  I'm inclined to unify the different log-levels to one single API call `bem_log` as I've proposed, and at the same time, keep the `_DEBUG` stuff around as it is now -- because that's important.

Is everyone OK with that, or would you rather I scrapped the whole idea and went with @SteveFosdick's original proposal?

hoglet67:
> So... back to my original point. I'm inclined to unify the different log-levels to one single API call bem_log as I've proposed, and at the same time, keep the _DEBUG stuff around as it is now -- because that's important.

I don't understand what this means in terms of the actual APIs we end up with.

I think I've already made my opinion clear.

ThomasAdam:
@hoglet67 -- one single call in terms of `bem_log` for all levels other than `LOG_DEBUG` as that will be `#ifdef`d out at compile time, as it is now for all other `_DEBUG` stuff.  So I'm hoping simple enough for most cases.

hoglet67:
So to be clear you are proposing:
```
enum log_level {
	LOG_INFO = 0,
	LOG_WARN,
	LOG_ERROR
};
void bem_log(enum log_level, const char *, ...);
void bem_log_debug(const char *, ...);
```
Honestly, I would still prefer:
```
void log_debug(const char *fmt, ...);
void log_info(const char *fmt, ...);
void log_warn(const char *fmt, ...);
void log_error(const char *fmt, ...);
void log_fatal(const char *fmt, ...);
```
as it's more uniform, allows debug to be dropped at compile time, and I think is simpler to use.

ThomasAdam:
@hoglet67 -- I'm not proposing that split, no.  `LOG_DEBUG` would still be part of the enum, it's just that you'd have to `#ifdef` out the specific calls to what would end up being `bem_log(LOG_DEBUG, ".......");`.  So the interface remains consistent with that we already have with regards to `_DEBUG` stuff.

hoglet67:
I'm now very confused.

Can you give an example of what might be included (e.g. in NS32016/32016.c) to log a message at debug level?

ThomasAdam:
```
#ifdef _DEBUG
bem_log(LOG_DEBUG, "This is only for debug: %s", something);
#endif

bem_log(LOG_WARN, "However, this is just for WARN...");
```

I use `bem_log` only for parity -- using `log` is probably easier.

hoglet67:
So each of those 177 debug log calls is now three lines rather than one? That doesn't seem the right trade off.

Also, using #ifdef like this we'll end up with lots of warnings of unused variables. I'm trying to get away from this. As both Steve and I have said before, the way to avoid that is an API that allows the following implementation to be provided in a header file.
```
static inline log_debug(const char *fmt, ...) {};
```
I'm pretty frustrated at the moment (and surprised) that we still seem to be going round in circles on this. It really doesn't bode well for anything that is more involved.

ThomasAdam:
So, unused variables, etc., where `_DEBUG` is concerned is already happening -- and that's the price at having to maintain two different states at **compile** time.

Note that `_DEBUG` is already established for just this purpose.  As for your example, that's **exactly** what I would have to do to circumvent the fact that as a performance hit, we couldn't possibly allow this at run-time, hence, at compile time, provide some empty alternative -- but in order to do that, this will require an API change in completely the wrong direction, IMO, i.e., to go back to the original idea of having separate `log_*()` functions.

You can't have it both ways, David.  Either you suck up the cost of using `_DEBUG` and actually litter the code base with those protections at compile time, or you don't, and instead do something altogether different.

I appreciate this is frustrating for you -- as it is for many, I'm sure.  Although I'm not intending to block this, the fact that I'm asking questions of all of this, and being met with resistance which is altogether curious, suggests this is still a good conversation to have.

However, it's clear from these discussions that I'm probably not the best person to carry  b-em forward.  I'm more than happy to pass the baton on to the next person, and I can merrily sit by and watch the code base become a convoluted mess.  If this is what has to happen, please let me know now, so I can cut my losses.

It's certainly been fun, if that's the case!  :)

hoglet67:
@ThomasAdam it was certainly not my intent to try to drive you away. 

I'm just frustrated because I feel an API that looks like:
```
void log_debug(const char *fmt, ...);
void log_info(const char *fmt, ...);
void log_warn(const char *fmt, ...);
void log_error(const char *fmt, ...);
void log_fatal(const char *fmt, ...);
```
is by far the best proposal, and I don't see any disadvantages.

Rather than making the code base a convoluted mess, this actually makes it cleaner. The vast majority of the code can use what ever log level suits, with no need to resort to #defines around each use of debug logging.

Providing a empty static inline implementation of log_debug() is subtly different to just #ifdefing out the method, as it avoids unused variable warning. If you don't believe me, then try the following example:
```
$ cat foo.c
#define f1(...)
static inline void f2(char *fmt, ...) {}
int main() {
       int x=123;
       int y=456;
       f1("test %d\n", x);
       f2("test %d\n", y);
}

$ gcc -Wall -ofoo foo.c
foo.c: In function 'main':
foo.c:6:12: warning: unused variable 'x' [-Wunused-variable]
        int x=123;
            ^
$
```

fordp2002:
I am confused why you object to lots of people with years of experience suggesting using tried and trusted design patterns is a bad thing. This is a collaborative open source project and a hobby project at that. I am happy to see people "scratching that itch" that they have.

ThomasAdam:
@fordp2002 -- you're absolutely correct.  I am not sure why I ever thought anything different would be a plausible alternative, let alone a kick in the teeth for those who were suggesting the original idea.  I can see now that I should have known better.

@SteveFosdick -- my sincere apologies, I will go ahead and merge your original implementation.

@hoglet67 / @fordp2002 / @SteveFosdick -- thanks, guys.  It's been fun.  Good luck with this project, but I'm done.

ThomasAdam:
@hoglet67 -- Thank you for your example, but it wasn't something I needed to understand the problem, and actually, the example there assumes there are things only applicable to `_DEBUG` which shouldn't be the case for the things we're referring to.  That is to say, by the time something for DEBUG logging is indicated, it's on the normal code-path.

startape:
Hi,

What happened here? I have come from the forums on suggestion of a few people. This was the first response I read. I am sad to understand what has happened here - how did it come to this?

I will try and submit my tape drive fixes - but now I am unsure how welcome they will be.

Having looked on the 'master' branch - is this a special branch? Only I see recently activity to fix a lot of things. Is this normal?

hoglet67:
Startape, your tape fixes will be most welcome.

Issue #20, opened 2017-03-08
i8271 emulation  always reports drive as ready


hoglet67:
See long discussion here:
http://stardot.org.uk/forums/viewtopic.php?f=4&t=12646

SteveFosdick:
So how does this fix it?  To emulate the hardware I would have expected there to be something counting disc_poll events after the end of an access.  Or does the DFS issue a command to deselect the drive and we assume the motor stops at that point?

In fact the spindown timer logic does seem to be in there, just not in your changes.

hoglet67:
@SteveFosdick -- it is there. The spindown function now clears the drive selects:
https://github.com/hoglet67/b-em/commit/9798bd8d3ee1966a693a37c2bc0607783565bb08#diff-dd4aee49713e76e2d448da1137e77fc1R91

SteveFosdick:
Ok, found it.  Another fine piece of work then.

SteveFosdick:
Ok, tested on Linux.  Checked that the DD noise was triogger by issuing a *CAT after it had gone quiet from the previous one - it was.  Made of copy of the welcome disk and used *BUILD to create a one-line text file.  Saved state and quit.  Started and loaded saved state.  Issued further *CAT file was there.  *TYPE of file gave correct contents.  This is with DFS 0.9 as per the B-Em distribution.

hoglet67:
Thanks for checking that. Would that sequence have previously failed?

SteveFosdick:
In the current stardot version which does not have your change applied it does seem to work fine for the most part - issuing a *. command causes the disk noise to start.  The one thing that does seem a little broken is upon loading the machine state from the file the initial *. command does not seem to cause any disk activity.  After that it goes back to working.

hoglet67:
The most serious effect of this bug was that you could end up with the catalogue of disk A being written onto disk B:
- In the Menu, LOAD Disk A
- *SAVE TESTA 4000 +100
- wait for drive to stop
- In the Menu, LOAD Disk B
- *SAVE TESTB 4000 +100

You'll now find that the catalogue of disk B has been replaced with that from disk A.

This doesn't happen with the fix, as long as you wait for the drive to stop running before changing disks.

SteveFosdick:
I did look to see if there was any interaction between disc loading and the controller emulations and didn't find any.  Perhaps the disc loader needs to call fdc_spindown if the new disc replaces an old one, i.e. the drive wasn't previously empty.

[...]

PR #22, opened 2017-03-10
8271 fixes to correctly emulate drive ready

hoglet67:
Fix for issue #20, plus a couple of additional suggestions from @SteveFosdick


ThomasAdam (approved):
Thanks, gents.  Looks good.

PR #25, opened 2017-03-17
Updated 32016 Co Pro to version with cpu_debug interface

hoglet67:
Also:
- Updated 32016 Co Pro to use logging API
- Some fixed to MinGW build


SteveFosdick:
Github gives me the option to "Rebase and merge" but I seem to remember from our previous discussions that this was part of the reason collaborating with git seemed to go wrong.  I am wondering if what would be better to do is "Create a merge commit" which I hope does exactly what "git pull" would do if it was a local branch.  I can't do that, though as it says in is no enabled on this repository.

hoglet67:
Try now, that option should be available.

SteveFosdick:
I think I have sorted the Windows issue.  Firstly the new CPU menu wasn't being called from the Windows version of debug_start and secondly at that point in the initialisation, drawing on the screen seems to cause a deadlock later on.  I think it may have been from attempting to nest two calls to startblit().

I have pushed the latest version to sf/debugging.

hoglet67:
Great, thanks. I can confirm that works for me now on Windows.

PR #26, opened 2017-10-03
Added Kieran's NULA support

hoglet67:
Change-Id: I3c7e33c167735a870ac5b58bdc881618035d6a8c

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,5 +23,4 @@ missing
 tags
 b-emlog.txt
 not_in_git/
-
-
+allegro.log
--- a/src/debugger.c
+++ b/src/debugger.c
@@ -644,6 +644,13 @@ void debugger_do(cpu_debug_t *cpu, uint32_t addr)
                         debug_outf("    Palette entries :\n");
                         debug_outf("     0=%01X   1=%01X   2=%01X   3=%01X   4=%01X   5=%01X   6=%01X   7=%01X\n", ula_palbak[0], ula_palbak[1], ula_palbak[2], ula_palbak[3], ula_palbak[4], ula_palbak[5], ula_palbak[6], ula_palbak[7]);
                         debug_outf("     8=%01X   9=%01X  10=%01X  11=%01X  12=%01X  13=%01X  14=%01X  15=%01X\n", ula_palbak[8], ula_palbak[9], ula_palbak[10], ula_palbak[11], ula_palbak[12], ula_palbak[13], ula_palbak[14], ula_palbak[15]);
+                        debug_outf("    NULA palette :\n");
+                        debug_outf("     0=%06X   1=%06X   2=%06X   3=%06X   4=%06X   5=%06X   6=%06X   7=%06X\n", nula_collook[0], nula_collook[1], nula_collook[2], nula_collook[3], nula_collook[4], nula_collook[5], nula_collook[6], nula_collook[7]);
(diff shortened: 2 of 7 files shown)


SteveFosdick (commented):
Dave, I have been out of the loop for a bit with work and family commitments but I think I just caught the NULA reaching a working state as hardware.

Assuming this is an option or is completely backwards compatible then I reckon it is better to commit it rather than have diverging forks.  Does it meet either of those criteria?

I am not saying we should have these in place before pulling this but it would be good to have some kind of video test suite even if it means comparing the output of a small set of programs with some reference screen shots.

hoglet67:
I believe NULA is backwards compatible, but best to ask Kieran and RobC.

kieranhj:
Hey folks, apologies for slow reply - just got back from a two week holiday.

The NULA implementation in b-em is backwards compatible insofar as the code effectively wrapped with "if nula_attribute_mode" for the extended MODE types and "if nula_palette_mode" for the colour palette indirection.

The largest and most impactful change is moving from an 8-bit palette framebuffer to a full 32-bit RGBA buffer for the emulator screen image composition.  To do this I replaced the direct writes to the framebuffer byte array with accessor functions for individual pixels.  In some cases there was a cast to write four black pixels at a time (0 as 32-bit integer) to the framebuffer.  This has been replaced by four separate pixel write function calls.

I have certainly noticed that the overall performance of the emulator is slower on my own PC.  Of course it still runs at full Beeb speed but an unscientific test running unrestricted (by holding Page Up) does not seem as fast as the original 2.2 version.

I do use a version of b-em with the NULA implementation as my development environment for PoP and have not noticed any other specific issues or unintended effects.

Note that this implementation of NULA is partial and not complete.  Specifically the pixel scroll and mask functionality is not implemented yet.  Perhaps I will do this at ABUG Nov.

SteveFosdick:
Kieran, did some CPU usage tests - just starting B-Em, selecting each display mode and watching the CPI usage and got some interesting results.  The effect is more noticable on an older, single-core laptop and on that, with the pre-NULA B-Em, I got CPU usage of about 25% in all the display modes.

With the NULA version the CPU usage is higher in all the modes but it also differs between modes:

Mode 0: 53%
Mode 1: 60%
Mode 3: 70%
Mode 4: 58%
Mode 5: 56%
Mode 6: 68%
Mode 7: 72%

Interestingly, the modes that need the most CPU are those where there are gaps between the lines, i.e. not all the output in the central block of the screen is coming from the BBC RAM but some of it is being generated independently, forced to black in the original ULA IIRC but presumably something else in NULA.

It's usable on the laptop, though I think it would hit the battery, and on my desktop which is a 4-core i5 the difference is only noticeably if you check the CPU usage - it doesn't feel any different.  I am thinking maybe merge this anyway but perhaps it would be worth then trying to see if we can make this run-time switchable.  Most of the files affected seem to be just a case of setting the bit-depth so this could just be a global variable that is either 8 or 32 depending on implementation.  I think that leaves pal.c and video.c so I am thinking we could have video-nula.c with your new NULA implementation, video-acorn.c with the original acorn one and video.c then becomes stuff that is common between the two with a video_init function that checks if NULA is enabled and then sets a few function pointers (for example videoula_write) to point to either the NULA or Acorn version.  It would also involve doing something similar with pal.c

PR #53, opened 2018-03-17
Allow normal typing in free run mode

ZornsLemma:
This is more of an RFC that a straight pull request but I thought this would be a convenient way for people to see the code.

This is a small change which temporarily disables free run mode while an emulated key is pressed down. The upshot of this is that you can toggle free run mode on by pressing SHIFT-PAGE UP, giving you a fast emulated machine, but you can still type normally without the emulated machine's OS seeing that each key is held down for a long time and triggering auto-repeat.

Without this patch, if I do SHIFT-PAGE UP and type "THE QUICK BROWN FOX JUMPS OVER THE LAZY DOG" I get something like "THEEE QUICCK BROOWN FOX  JUMPPS OVER THE LAZY DOOGG"; with this patch, I can type normally even after SHIFT-PAGE UP.

I think this is useful for running non-game applications (compilers, text editors, etc), particularly when using the emulator to do development of new application on.

It might be worth supporting this when using speeds other than 100%, but I haven't looked into that yet  - currently it only affects free run mode.

### DIFF
--- a/src/b-em.h
+++ b/src/b-em.h
@@ -33,6 +33,7 @@ extern char exedir[512];
 extern int joybutton[2];
 
 extern int bempause;
+extern int bemfreerun;
 
 void setquit();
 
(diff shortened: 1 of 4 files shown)


SteveFosdick:
This works well.

PR #62, opened 2018-05-26
Minor changes to get b-em/sf/allegro5 to build on latest macOS

myelin:
- Switch to gnu11 from gnu99 to make static_assert work, and get rid of a _static_assert symbol not found error when linking.
- Add -lallegro_main when linking, to get rid of a _main symbol not found error.


SteveFosdick (approved):
Still compiles on Linux with these changes and Windows has its own Makefile.

PR #97, opened 2020-02-29
Fixes to allow build on MSYS2

dominicbeesley:
Hello,

I've been trying to build on MINGW64 / MSYS2 and I have the attached fixes which give a working executable that look to be safe enough to include in master. I don't have access to my Linux box to double check the build though

- win.c was not being included in the default makefile
- detection for type max_align_t was broken

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,3 +34,11 @@ src/gtest.exe
 src/fsmap
 src/hdfmt
 src/fdresize
+packages/*
+src/Debug/*
+src/Release/*
+b-em.iobj
+b-em.ipdb
+b-em.iobj
+b-em.pdb
+src/hdfmt.exe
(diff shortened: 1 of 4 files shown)


SteveFosdick:
Thanks, dominic.  It still builds correctly and at least starts on Linux.  Cross-compiling for Windows still seems to work too though I just want to check that the resulting executable runs.  I'll wait until my Windows VM has finished doing Windows Update.

dominicbeesley:
Thanks steve,

I've also got some further fixes - the final copy step needed a tweak for
the .exe extension and I've got an updated visual studio project. Would you
like me to send them through as separate pulls?

I've also got a Paula emulation (see
https://stardot.org.uk/forums/viewtopic.php?f=53&t=18340) would you be
interested in this going in the main codebase or should i keep that in my
own fork?

Cheers

Dom

SteveFosdick (approved):
No problems found - still compiles on Linux and cross compilers on Linux for Windows and resulting executable run on each.

PR #98, opened 2020-02-29
fix dodgy mode 7 separated graphics glyphs

dominicbeesley:
Some of the glyphs were slightly wrong for separated graphics


SteveFosdick (approved):
I will have to trust you on the accuracy of this one.  It looks fine to my eyes.

PR #99, opened 2020-03-02
Vsbuild

dominicbeesley:
This pull should allow building in Visual Studio 2017 - the old vs project was quite out of date.


fordp2002:
Maybe add a wiki page with how to build in VS2017?

dominicbeesley:
> Maybe add a wiki page with how to build in VS2017?

I did consider it but It should* be a case of opening the corrected project in VS and hitting build. The nuget package manager should then update the packages if they need it and it should just work. 

* with VS you can never tell what new feature will come along and b****er it all up though!

SteveFosdick (approved):
This compiles cleanly on Linux and cross-compile for Windows (from Linux) and runs from both including 32016 2nd proc.

PR #100, opened 2020-03-11
vsprojx fix for 32016 and a debugger bugfix

dominicbeesley:
A define was missing for the 32016 and I spotted that the debugger if given an address >&FFFF was crashing

Not sure how to easily split this into two pull requests!

D

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,7 @@ autom4te.cache/
 b-em
 b-em.exe
 b-em.res
+b-em.ilk
 config.log
 config.status
 configure
@@ -42,3 +43,5 @@ b-em.ipdb
 b-em.iobj
 b-em.pdb
 src/hdfmt.exe
+scsi/*
+
--- a/src/6502.c
+++ b/src/6502.c
@@ -292,11 +292,13 @@ static inline uint16_t read_zp_indirect(uint16_t zp)
 static uint32_t do_readmem(uint32_t addr)
 {
 
-        if (addr < 65536)
-            if (pc == addr)
-                fetchc[addr] = 31;
-            else
-                readc[addr] = 31;
+        if (addr >= 0x10000)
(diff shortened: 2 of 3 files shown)


fordp2002:
I prefer the constant 0x10000 to 65536. Possibly better to change the code so you can use 0xFFFF which has even more meaning.

dominicbeesley:
> I prefer the constant 0x10000 to 65536. Possibly better to change the code so you can use 0xFFFF which has even more meaning.

I wasn't sure, so I followed the style where the arrays are defined in 6502.h and elsewhere in the code.

fordp2002:
No problem, just a personal preference.

dominicbeesley:
Ah, hang on probably better hold off on this and I'll have a more in-depth look in the morning, there's other stuff that fails when address is > &FFFF. I'll come up with a better fix tomorrow.

Would it be better to just make read/writemem take a uint16_t? If not then what should the default out of bounds value be 0xFF, high 8 bits of address?

D

dominicbeesley:
I've now updated this pull to just ignore reads/writes above 64k which seems safe and logical

SteveFosdick (commented):
This latest version looks fine.  To answer the question about maybe using a unit16_t for the address, the reason it's uint32_t is because they debugger is intended to be useful for all the processors and there is a set of these functions for the various processors that must all use the same calling convention.

SteveFosdick (changes requested):
The only barrier I can see to merging this is that github says there are merge conflicts for the MS Visual C++ changes.  As that's your area of expertise and not mine it would be better you resolve them than if I try to do so.

dominicbeesley:
Thanks Steve,

I've done the resolve conflicts, should be good to go

D

SteveFosdick (approved):

PR #106, opened 2020-06-12
Full screen fixes

dominicbeesley:
ALT-Tab away from b-em and back to b-em would cause the Allegro key state to get stuck as if alt pressed. Maintain own alt flag and reset on lose focus

Fullscreen switch didn't work first time (appears to be an Allegro problem, force a second switch)

Restore window size when switch back from full screen

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -68,6 +68,7 @@ int autoboot=0;
 int joybutton[2];
 float joyaxes[4];
 int emuspeed = 4;
+bool alt_down = false;
 
 static ALLEGRO_TIMER *timer;
 static ALLEGRO_EVENT_QUEUE *queue;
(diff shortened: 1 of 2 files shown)


SteveFosdick (approved):

SteveFosdick:
BTW, looking through your fullscreen because of issues compiling with GCC 10, I noticed that in savestate.c for the Paul sound you haven't added a call to paula_loadstate from within load_state_two.

load_state_one is the legacy function in which a snapshot file has the sections in a fixed order and all must be present, such that snapshot files can only be loaded into a version of b-em that emulated the same set of hardware.

load_state_two will load the sections in any order, skipping sections it doesn't recognise and just not reloading the state of any hardware the current emulator emulates but wasn't there when the snapshot was saved.

dominicbeesley:
> BTW, looking through your fullscreen because of issues compiling with GCC 10, I noticed that in savestate.c for the Paul sound you haven't added a call to paula_loadstate from within load_state_two.

Thanks I'll check my other code for this too - I have to admit I didn't get my head into the load/save state thing adequately I just cribbed what was in music5000 but didn't test it properly. I've got some free time today so I'll try and get time to properly check

PR #121, opened 2020-08-28
Set Allegro VSYNC to off

ajgbarnes:
Call al_set_new_display_option to suppress vsync to improve performance

Without this on an i7 3GHz machine on Ubuntu 20.04 with an NVidia card I only get 60% / 1.2 MHz at 100%
With this display option set this increases to 100% / 2 Mhz

Tested with White Light and no screen tearing apparent at all on Linux or RPi.

Note that this can be overriden by the graphics driver or sometimes not supported - for the latter case ALLEGRO_SUGGEST will still allow the display to be created.


SteveFosdick (approved):
This works fine on Windows 10 on Virtualbox and on my ArchLinux machine with Intel integrated graphics.

PR #123, opened 2020-09-24
Jy/hfe

jamesyoungman:
This is an initial implementation of HFEv3 disc image support for B-EM.   Only HFE versions 1 and 3 are supported.  Please note some limitations of the current code:

- It's read-only.   The writesector and format operations don't modify anything.
- Write operations in Watford DFS don't correctly fail (see https://stardot.org.uk/forums/viewtopic.php?f=4&t=20601 for details)
- Double-density and double-sided support is planned and part-implemented, but not tested.

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -63,6 +63,7 @@ b_em_SOURCES = \
 	disc.c fdi.c \
 	fdi2raw.c \
 	gui-allegro.c\
+	hfe.c \
 	i8271.c \
 	ide.c \
 	joystick.c \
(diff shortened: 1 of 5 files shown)


SteveFosdick (approved):
Thanks for doing this.  Even without it working on Watford/Opus this will be useful so I will merge it.

SteveFosdick:
James, do you have an HFE file you tested this with?  When I try it I get CRC data errors on every sector.

jamesyoungman:
I do.  I will attach it on the Stardot thread.   But, I would like to use your test file too, to reproduce the problem.

PR #126, opened 2020-10-05
Compiling version of B-em on Visual Studio 2019

hfmanson:
- exclude linux source files from VS project
- v141 -> v142
- re-include pdp11 tube files
- re-include led.*, hfe.*, and imd.* files
- Win32 implementation of ftruncate
- Excluding 68000 tube which can be turned on by #define M68K
- Not using sdf_lock functions on Win32

### DIFF
--- a/src/b-em.vcxproj
+++ b/src/b-em.vcxproj
@@ -14,21 +14,21 @@
     <ProjectGuid>{28E2DE55-0A88-47FA-92DC-3F96D72608F9}</ProjectGuid>
     <Keyword>Win32Proj</Keyword>
     <RootNamespace>bem</RootNamespace>
-    <WindowsTargetPlatformVersion>10.0.17763.0</WindowsTargetPlatformVersion>
+    <WindowsTargetPlatformVersion>10.0</WindowsTargetPlatformVersion>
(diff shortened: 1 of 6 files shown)


SteveFosdick (changes requested):
Thanks for taking the time to make this compile in Visual Studio again.

Can you look again at sdf-acc.c, please.  I thought I had that so the sdf_lock function has conditional assembly, i.e. it should compile to an empty function on Windows.  A modern compiler should be clever enough to then omit the call too as it has static scope.  Also the fflush at spindown is to make sure any other programs on the host see a consistent SSD file so should probably not be subject to conditional assembly.

SteveFosdick (approved):
On second thoughts, I can see why you have put the conditional assembly where you did - the constants F_WRLCK and F_UNLCK and presumably not defined on Windows.  I'll merge this but tidy up the conditional compilation in this module a bit further.  Can you then see if it compiles in VC still, please.

SteveFosdick:
Ok, I have pushed a change to master.  Could you check it still compiles on VS, please.

hfmanson:
It still compiles on VS

Issue #140, opened 2021-03-13
Partly broken build on macOS


lurkio:
Some strange workarounds are needed if you want to build B-em on macOS Mojave or Catalina:

https://stardot.org.uk/forums/viewtopic.php?p=312835#p312835

The *gtest* and *sdf2imd* directories are empty. What are we missing out on?!

SteveFosdick:
See: https://stardot.org.uk/forums/viewtopic.php?p=312952#p312952
Conditional compilation for unlocked I/O in https://github.com/stardot/b-em/tree/sf/issue140, commit https://github.com/stardot/b-em/commit/84d3a464f30153ed7b1f0d3664d24d2e1e9c338c

SteveFosdick:
Thanks, Steve, all changes merged.

PR #141, opened 2021-03-15
Further partial fix for Issue #140 building on OS X

sai2791:
I only meant to create the pull request for configure.ac, but the copy of master has a further change for your issue 138.  Unfortunately I do not know how to remove this second change.

The error: expected expression is corrected by specifying c++11, since the CFLAGS use gnu I changed to the -std=gnu++11 for the c++ compiler.

The issues with gtest and sdf2imd I am still looking at, and will create a final pull request once I get it figured out.


sai2791:
I think I have royally screwed this pull request and will submit another

PR #142, opened 2021-03-15
Another Partial fix for Issue #140

sai2791:
Partial fix for Issue #140.  This removes the error: expression expected in debugger_symbols.cpp

debugger_symbols.cpp:32:16: error: expected expression
    map.insert({ e->getSymbol(), e });
               ^

The remaining issue with gtest and sdf2imd I am still looking into to see if I can find a proper fix.


SteveFosdick:
Thanks, Steve.

PR #166, opened 2022-03-04
Hide Menu in Fullscreen mode

Homo-carbonis:
I've written a function to hide the menu bar when in fullscreen mode. It's very useful on Linux because a problem with allegro causes the program to get stuck if you open a dialogue box while in fullscreen but if other people are able to use the menus normally in fullscreen mode it might be better to have a separate keyboard shortcut to toggle menu visibility. Please let me know what you think.


SteveFosdick:
Sorry it has taken me to long to get around to looking at this.  Trying your version it seems to work as described.  There is something a little strange, though, in that when I hit the key to make b-em go full-screen the existing window seems to expand, as if to go full-screen, as a kind of animation, then shinks again to normal size.  Finally, it goes full-screen after a tiny delay.

Was there intended to be some animation here, or is it doing something that is perhaps unecessary and could be removed?

Homo-carbonis:
My code just deletes the menu and then goes fullscreen as normal. I don't get the same effect but I think it must be coming from allegro.

SteveFosdick (approved):
I'll merge this.  I do have a small change to resolve the animation issue I was seeing but I will apply this afterwards.

SteveFosdick:
I have now added a further commit to the master branch that resolves the strange effect for me.  Essentially I reverse the order of removing the menu and going full-screen for the enter full-screen case.

I did note that I didn't have an issue with stuck dialog boxes even in full-screen mode.  My test case was the file chooser dialog to select a disc image, i.e. Disc->Load Disc 0/2

Another possibilty is to change the behaviour according to how you enter full-screen.  If you do so via the keyboard, and therefore obviously know the key concerned, then remove the menu.  If you enter full-screen from the settings menu then leave the menu so you can use it to get out of full-screen again as well as any of the other menu options.

BTW, you can re-map what keys do the various things such as full-screen via Settings->Keyboard->Remap Keyboard.  Once the picture of the keyboard appears, you click one of the green buttons that don't exist on a real keyboard and represent emulator-specific actions, then hit the PC key you want to connect it to.  It should remember the ALT state too.

Homo-carbonis:
I think the freezing issue is specific to XWayland, or possibly even to the my window manger. There's an Allegro bug report about issues with fullscreen mode which might have something to do with it.

I did think about only hiding the menu if you enter fullscreen from the keyboard but didn't get around to it. I think it's a good idea so I'll do that next.

PR #167, opened 2022-03-16
Only hide menu when going fullscreen from keyboard

Homo-carbonis:
I've added a different function to keep the menu bar when you select fullscreen from the video menu.


SteveFosdick (approved):
This works well for me.

PR #188, opened 2023-02-18
Add experimental support for emulation of the Sprow ARM7TDMI co-processor.

rjpontefract:
Add experimental support for emulation of the Sprow ARM7TDMI co-processor.
This was ported from BeebEm so credit for the code goes to the BeebEm team.
Many thanks to Robert Sprowson for permission to include the firmware with b-em.


SteveFosdick (approved):

SteveFosdick:
I have merged this but when I built the merged version I got some warnings from gcc.  The one about comments seems harmless enough but the other two looked significant:

```
sprow.c: In function ‘sprow_dbg_disassemble’:
sprow.c:273:3: warning: ‘strncpy’ output truncated before terminating nul copying as many bytes from a string as its length [-Wstringop-truncation]
  273 |   strncpy(buf, dest, strlen(dest));
      |   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
and
```
ARMulator/armemu.c:841:24: warning: ‘do_int’ may be used uninitialized [-Wmaybe-uninitialized]
  841 |                     if (do_int && (cp14r0 & ARMul_CP14_R0_INTEN2))
      |                        ^
ARMulator/armemu.c:816:25: note: ‘do_int’ was declared here
  816 |                     int do_int = 0;
      |                         ^~~~~~
```
I have pushed fixes in commit https://github.com/stardot/b-em/commit/58e71860a7db660013b16130901046f9a6e62102  Could you check the one where I haved moved the initialisation of do_int.  The issue here is that there is a goto into that block where the labels is after the initialisation which gcc seems to interpret as the initialisation not being done.  Moving if after the label silences the warning.

rjpontefract:
Those changes look fine to me.  I built it with clang on macOS and got a different set of warnings.  I'll build it with GCC and see if there's anything else that needs attention.

PR #189, opened 2023-02-19
Fix build warnings when building with GCC on Linux

rjpontefract:
I've fixed the other warnings you were probably seeing when you built on Linux with GCC.


SteveFosdick (approved):

rjpontefract:
I didn't include the changes you  made in sf/sprowfix for do_int and strncpy(), so please could you merge those changes into master as well?

PR #190, opened 2023-02-23
Rp/clang warnings

rjpontefract:
I fixed all of the warnings generated on macOS Ventura 13.2.1 with Apple clang version 14.0.0 (clang-1400.0.29.202).  

I also built on Linux gcc (Ubuntu 11.3.0-1ubuntu1~22.04) 11.3.0 and no warnings were generated there either.

I haven't built this on Windows.  Please can you let me know which compiler/dev environment you use on Windows so that I can set up the same?

### DIFF
--- a/src/ARMulator/armemu.c
+++ b/src/ARMulator/armemu.c
@@ -783,6 +783,7 @@ ARMul_Emulate26 (ARMul_State * state)
         {
             ARMword cp14r0;
             int ok;
+            int do_int = 0;
 
             ok = state->CPRead[14] (state, 0, & cp14r0);
 
@@ -816,7 +817,6 @@ ARMul_Emulate26 (ARMul_State * state)
 
                     state->CP14R0_CCD = (ARMword)-1;
 check_PMUintr:
-                    int do_int = 0;
                     cp14r0 |= ARMul_CP14_R0_FLAG2;
                     (void) state->CPWrite[14] (state, 0, cp14r0);
 
--- a/src/hdfmt.c
+++ b/src/hdfmt.c
@@ -101,7 +101,8 @@ int main(int argc, char **argv)
 {
         const char *fn;
         int status, len, size, sectors, cyl;
-        char *dat_fn, *dsc_fn, geom[22];
+        char *dat_fn, *dsc_fn;
+        unsigned char geom[22];
         FILE *dat_fp, *dsc_fp;
 
         if (argc == 3)
--- a/src/mmccard.c
+++ b/src/mmccard.c
@@ -84,7 +84,7 @@ void mmccard_write(uint8_t byte)
                             break;
                         case 0x50: // Set block length.
                             address = (mmc_args[0] << 24) | (mmc_args[1] << 16) | (mmc_args[2] << 8) | mmc_args[3];
-                            log_debug("mmccard: set block length to %lx", address);
+                            log_debug("mmccard: set block length to %llx", address);
                             if (address > 0 && address <= sizeof(mmc_buffer)) {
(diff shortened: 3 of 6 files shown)


SteveFosdick (commented):
Thanks for this merge request.  It is also great to hear of b-em being compiled on a Mac.

I think this one will need a little bit more checking.

At a first glance, commit 88b411d is reverting a previous commit that was itself introduced to fix an uninitialised variable warning from gcc.  I think the syntax where a combined variable declaration and initialisation can appear anywhere, and not just at the start of a block, is down to the version of the C standard and compilers sometimes have options to say which one they should accept.  If clang doesn't support that syntax we will have to split the declaration and the initialisation.

On the width of an off_t, most modern systems would have this as 64bit, including some 32 bit versions of Linux, though some may stick to 32 bit.  To make it more complicated, the width of long also varies between systems.  We also need to be careful about not introducing printf conversion characters not supported by Windows and, in that respect, Microsoft's Windows compiler is a C++ compiler that makes no effort to support anything from a C standard that is not also C++.

Then some of the commits are fixing obvious bugs.  How assigning the result of fopen to an int didn't generate a warning before I don't know.

On the subject of building for Windows, there are at least three ways.  I cross-compile it from Linux using the script mkcross.sh which relies on having gcc installed as a cross-compiler with a particular (common) naming convention.
`
It is also possible to build it on Windows using MingW.  This can be quite tricky to set up so I could share a ZIP if that helps.  Finally, some people build it on Windows with Visual Studio.  That may invove fixing more warnings or even errors due to compiler quirks.

rjpontefract:
The problem with "int do_int = 0;" was due to clang not allowing a declaration directly after a label, whereas gcc does seem to allow it (the C standard doesn't allow it, so clang is arguably correct here).  So the previous commit where it was moved after the "check_PMUintr:" label broke it with clang.

The original position of that code was inside the else block containing the label, hence the warning that the goto skipped the initialisation.  Moving it to the containing block as it is now puts it in a scope where it is initialised regardless of the goto.  We could move it here:

```
check_PMUintr:
                    cp14r0 |= ARMul_CP14_R0_FLAG2;
                    int do_int = 0;
```
so that a statement follows the label, or we could introduce an empty statement after the label like this:

```
check_PMUintr:
                    ;
                    int do_int = 0;
                    cp14r0 |= ARMul_CP14_R0_FLAG2;
```
The code in the current pull request builds warning free with both clang and gcc however, but one of the above might be neater.

SteveFosdick (approved):
Ok, on do_int, I didn't see that you'd moved the scope - the graphical diff doesn't show much context.  I'll merge this and see how well it builds for Windows.

rjpontefract:
The off_t format specifier is a tricky one.  On Linux with gcc it's defined as a long int:

`typedef long int __off_t;`

whereas on macos with clang it's defined as a __int64_t:
```
typedef __int64_t __darwin_off_t;
typedef __darwin_off_t off_t;
```
Both are 64 bit values but gcc needs %lx or %ld and clang needs %lld or %llx as the format specifier.
Using %j and casting to intmax_t is portable but does require a C99 compliant compiler.

rjpontefract:
I have another branch called rp/windows_fixes where I have changes that allow it to build and run with Visual Studio 2017.
There are a lot of warnings from the compiler but no errors.  I can fix some of the warnings in this branch if it would help.
I haven't built it with mkcross.sh yet.  I do have the cross compiler installed on Linux but I haven't built a cross compiled version of the allegro library yet.  Any hints on how you did that would be appreciated.  Also, the zip file for MingW would be appreciated.

PR #192, opened 2023-03-15
Add support for the Watford Electronics ROM/RAM board.

rjpontefract:
This small set of changes adds support for the Watford Electronics RAM/ROM board.  The board has 128K of DRAM and 16K SRAM on it along with ROM sockets.  The 128K of DRAM can be used as 8 SWR banks 0 through 7 or, using ROMs provided with the board, a 128K RAM disk or printer buffer.
The 16K SRAM is available as SWR bank 14.
Writing to the DRAM or SRAM is controlled by a latch controlled by writes to address &FF30 to &FF3F.
I added this to b-em with a new SWRAM configuration type of weramrom which configures the RAM banks, ROM slots and invokes the code to handle the latch and writes to &8000 to &BFFF.  Is this the right way to do it?
I've built and tested it on Linux, Windows and macOS.  Would it help if I included an image of the disk that comes with the board and contains the silicon file system and printer buffer ROMS?

### DIFF
--- a/b-em.cfg
+++ b/b-em.cfg
@@ -406,6 +406,20 @@ rom15=integra120
 rom14=basic2
 rom13=vdfs
 rom12=dfs09
+[model_27]
+name=BBC B w/1770 FDC + WE ROM/RAM
+fdc=acorn
+65c02=false
+b+=false
+master=false
+modela=false
+os01=false
+compact=false
+os=os12
+tube=none
+romsetup=weramrom
+rom15=basic2
+rom13=dfs226
 [disc]
 defaultwriteprotect=false
 scsienable=false
--- a/src/6502.c
+++ b/src/6502.c
@@ -745,11 +745,34 @@ static void write_fe34(int val)
 
 static void do_writemem(uint32_t addr, uint32_t val)
 {
-        int c;
+    int c;
+    static int watford_ram_bank = 14;
 
     addr &= 0xffff;
 
         writec[addr] = 31;
+
+        // Watford ROM/RAM board
+        if (!strcmp(models[curmodel].romsetup->name, "weramrom"))
+        {
+          if (addr >= 0xFF30 && addr <= 0xFF3F)
+          {
+            watford_ram_bank = addr - 0xFF30;
+            return;
+          }
+
+          if (addr >= 0x8000 && addr <= 0xBFFF)
+          {
+            int memtype = rom_slots[watford_ram_bank].swram ? 1 :2;
+            romsel = watford_ram_bank << 14;
+            if (memtype == 1)
+            {
+              uint8_t *base = rom + romsel - 0x8000;
+              *(base + addr) =  (uint8_t)val;
+            }
+            return;
+          }
+        }
 
         c = memstat[vis20k][addr >> 8];
         if (c == 1) {
--- a/src/mem.c
+++ b/src/mem.c
@@ -272,6 +272,25 @@ void mem_romsetup_master(void) {
     exit(1);
 }
 
+void mem_romsetup_weramrom(void) {
+    const char *sect = models[curmodel].cfgsect;
+    int slot;
+
+    load_os_rom(sect);
+    for (slot = 15; slot >= 0; slot--)
+        cfg_load_rom(slot, sect);
+
+    rom_slots[14].swram = 1;
+    rom_slots[7].swram = 1;
+    rom_slots[6].swram = 1;
+    rom_slots[5].swram = 1;
+    rom_slots[4].swram = 1;
+    rom_slots[3].swram = 1;
+    rom_slots[2].swram = 1;
+    rom_slots[1].swram = 1;
+    rom_slots[0].swram = 1;
+}
+
 int mem_findswram(int n) {
     int c;
 
--- a/src/mem.h
+++ b/src/mem.h
@@ -22,6 +22,7 @@ extern void mem_romsetup_swram(void);
 extern void mem_romsetup_bp128(void);
 extern void mem_romsetup_master(void);
 extern void mem_romsetup_compact(void);
+extern void mem_romsetup_weramrom(void);
 extern void mem_fillswram(void);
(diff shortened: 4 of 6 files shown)


SteveFosdick:
I tried merging this into a local branch of mine and then made the mistake of pushing it to master so Github shows this as merged.  At that stage I still wanted to review re-work part of it so I have reverted those changes on master.

So to clarify the expected behaviour, when writing to the &FE30-&FE3F region, the ROM bank for reading is controlled by the value written whereas the bank for writing to is controlled by the lowest four bits of the address.

I have written a mixed BASIC/Assembler test program to test this which I include below.

I have re-worked this a bit to better fit with the way B-Em does memory and paged ROMs generally and the result is in the branch https://github.com/stardot/b-em/tree/sf/weramrom  Essentially, to keep memory access efficient B-Em uses a set of tables memstat[] and memlook.  These are indexed by the state of shadow memory switching and the upper eight bits of the address.  memstat returns a memory type.  The values before this change were:

0 - I/O
1 - RAM
2 - ROM

This was all Sarah Walkers code.  Then memlook contains pointers so a memory access can then be completed by indexing these pointers by the lower 8 bits of the address, unless it's I/O in which case a big case statement is used.  Normal ROM paging works by changing the pointers in memlook for the region concerned and updating memstat according to whether the bank is ROM or RAM.

That mechanism doesn't support split read/write so I have added a new memory type - 3 - for the cases where the write bank is not the same as the read bank and introduced weramrom_base.  It this split bank arrangement the base pointers in memlook are correct for the read bank so the do_readmem needs no change.  The do_writemem function processes memory type 3 by using weramrom_base instead of memlook.  This could be extended to other sideways RAM systems that have a separate write bank.

You mentioned bank 14 is static RAM and the first 8 are dynamic RAM.  Does the board have battery backup for bank 14?  

It probably would be useful to have the original software too for completeness.

Anyway, the test program:
```
   10 REM > WERTEST
   20 REM Watford ROM/RAM Board Test
   30 DIM C% 256
   40 FOR N%=0 TO 3 STEP 3
   50   P%=C%
   60   [OPT N%
   70   PHP
   80   SEI
   90   LDA &F4
  100   PHA
  110   TYA
  120   STA &FE30,X
  130   LDA &70
  140   STA &8006
  150   CMP &8006
  160   BEQ ok
  170   LDA #&00
  180   BEQ both
  190   .ok
  200   LDA #&01
  210   .both
  220   STA &72
  230   TXA
  240   STA &FE30,X
  250   LDA &8006
  260   CMP &70
  270   BNE notok
  280   LDA &72
  290   OR A #&02
  300   STA &72
  310   .notok
  320   PLA
  330   STA &F4
  340   STA &FE30
  350   PLP
  360   RTS
  370   ]
  380 NEXT
  390 PRINT ~C%;", press any key..."
  400 G%=GET
  410 FOR Y%=0 TO 15
  420   FOR X%=0 TO 7
  430     PROCisram
  440   NEXT
  450   FOR X%=8 TO 13
  460     PROCnotram
  470   NEXT
  480   X%=14
  490   PROCisram
  500   X%=15
  510   PROCnotram
  520 NEXT
  530 END
  540 DEF PROCisram
  550 PROCtest
  560 IF X%=Y% AND (?&72 AND 1) = 0 THEN PROCerror("mismatch on same bank")
  570 IF X%<>Y% AND (?&72 AND 1) = 1 THEN PROCerror("no write differentiation")
  580 IF (?&72 AND 2) = 0 THEN PROCerror("read-back failed")
  590 ENDPROC
  600 DEF PROCnotram
  610 PROCtest
  620 IF ?&72 THEN PROCerror("non RAM bank writable")
  630 ENDPROC
  640 DEF PROCtest
  650 ?&70=RND(256)-1
  660 ?&71=RND(256)-1
  670 CALL C%
  680 ENDPROC
  690 DEF PROCerror(E$)
  700 PRINT E$
  710 PRINT"  with read bank ";Y%
  720 PRINT"  and write bank ";X%
  730 ENDPROC
```

SteveFosdick:
Oh, some comments on the test program.  Y% is the read bank, X% the write bank.  Note the wait for a keystroke - it is so I can set a breakpoint in the debugger to the assembler bit if necessary.

rjpontefract:
Hi, thanks for taking a look

_So to clarify the expected behaviour, when writing to the &FE30-&FE3F region, the ROM bank for reading is controlled by the value written whereas the bank for writing to is controlled by the lowest four bits of the address._

Writing is controlled by writing to &FF30-&FF3F not &FE30.   The value written isn't important, just the address.
Reading is controlled as normal using the value written to &FE30.

_You mentioned bank 14 is static RAM and the first 8 are dynamic RAM. Does the board have battery backup for bank 14?_

Yes, bank 14 does have battery backup and the real board has write enable/disable and read enable/disable switches for that bank.

_It probably would be useful to have the original software too for completeness._

I've attached a disk image of the original software.  This includes ROMs for the Silicon File System or a printer buffer that can be loaded in bank 14 and use the first 8 DRAM banks for storage.  

For completeness, I've also included the manual for the board which explains the bank selection.

Hope that helps.

[Watford Electronics ROM RAM.ssd.zip](https://github.com/stardot/b-em/files/11009827/Watford.Electronics.ROM.RAM.ssd.zip)

[Watford Electronics RAM ROM Board.pdf](https://github.com/stardot/b-em/files/11009829/Watford.Electronics.RAM.ROM.Board.pdf)

SteveFosdick:
Ok, it seems I misunderstood slightly because FE30 and FF30 look very alike.  I have pushed further changes to the https://github.com/stardot/b-em/tree/sf/weramrom branch to correct that and also to implement the battery backup on bank 14.  It passes the TEST program on the disc above.

rjpontefract:
Easy to overlook, especially as FE30 is what you're expecting it to be.
I checked out the branch and it looks OK to me.  The tests ran OK and the SFS worked OK.
Thanks for reworking my code.

PR #208, opened 2023-07-23
sound: Poll SN76489 at 125 kHz

sarah-walker-pcem:
Polling the SN76489 at 125 kHz (the native "update" rate on real hardware, 4 MHz / 32) improves the audio quality of multi-voice sample playback, by significantly reducing the time spent in "in-between" states where not all voices have yet had their volumes updated.

As part of this, the handling of SN write is modified to a poll as part of the 125 kHz update routine. This allows multiple writes with the chip enable held active to be honoured (eg the Castlevania III music conversion).

### DIFF
--- a/src/6502.c
+++ b/src/6502.c
@@ -235,6 +235,7 @@ static inline void polltime(int c)
     via_poll(&sysvia, c);
     via_poll(&uservia, c);
     video_poll(c, 1);
+    sound_poll(c);
     otherstuffcount -= c;
     if (motoron) {
         if (fdc_time) {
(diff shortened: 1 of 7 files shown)


SteveFosdick (approved):

rjpontefract:
Thanks for a nice update for the sound code, the speech demo is much better than it was before.

One thing I noticed was that on macOS there's the following build warning:

sysvia.c:87:17: warning: unused variable 'oldIC32' [-Wunused-variable]
        uint8_t oldIC32 = IC32;

SteveFosdick:
I have pushed a further commit to fix this warning.  The compiler is right, the variable was unused.

PR #255, opened 2026-01-04
Normalise whitespace

fesh0r:
This is just a whitespace cleanup across all files.

I've removed trailing whitespace and extra newlines at the end of files.

Also have converted tabs to spaces in all files except Makefiles. This was done on a per file basis using the number of spaces that matched the existing leading indentation and aligned comments in the file, which usually ended up being 8 spaces.

I have not re-indented or re-formatted any files in these commits, so some files such as 6502.c still have inconsistent indenting and formatting within the file itself due to being modified over time by different people


fesh0r:
Is this likely to be merged or is it a bit too disruptive?
I have a bunch of build patches that are currently on top of this, but will rebase if needed.
Main reason for this was a lot of editors trim trailing spaces by default these days so easy to get a lot of noise in diffs.
Also due to the editorconfig file get lots of misaligned comments due to embedded tabs in the older source

SteveFosdick:
Sorry it has taken me so long to respond to this.

I know opinions are divided on reformatting but it seems to me to be swings and roundabouts.  Reformatting means comparing two commits that are long time apart (across the reformat) can show lots of extra noise but I have also been caught with things like an editor removing trailing spaces and thus a recent commit shows extra noise, or converting tabs to the wrong width and maybe that is worse.  I do also like to have things look sensible, with consistent indentation.  So, in principle I'm fine with this.

One thought, though - where do we expect changes to modules that have been copied in from somewhere else, like musahi to come from?  If we want to track the module from wherever it is being developed, reformatting may make that harder though, if it can be done with some kind of 'pretty printer' then we do that on incoming files to make it easier.  If we will maintain it ourselves, then maybe being consistent with the rest of b-em would be better.  What do you think?

fesh0r:
I think I'll drop this cleanup for now. While limiting it to b-em origin code is doable this will still make merging things like the ongoing tape overhaul harder in future, and was a bit of a blunt fix to something that could be worked around just by changing my vscode config. Can always come back to it later when there is less development happening outside of this repo.

On your other note, I have been trying to track down the upstream for the external modules, but not having much luck with some, will keep digging though for documentation purposes if nothing else.

I'll rebase my unix build updates branch without this cleanup and send that through for you to have a look at

PR #262, opened 2026-09-04
NS32016: add T-bit trace trap support to permit single-stepping instructions

bson:
Add T-bit trace trap support to permit single-stepping instructions in a debugger.  Like with the actual NS32016 the T-bit does nothing until explicitly enabled.

I'm using an emulator derived from this in b-em to implement a gdbserver for NS32016, to test it out before deploying and testing against real hardware.  I had to add this for it to actually function since it's central to any debugger.  I realize none of your current targets would likely use it for anything, but since it doesn't do anything unless enabled it can't hurt, and it does make the emulation ever so slightly more complete.

### DIFF
--- a/src/NS32016/32016.c
+++ b/src/NS32016/32016.c
@@ -3068,5 +3068,10 @@ void n32016_exec()
          }
       }
 #endif
+
+      if (T_FLAG)
+      {
+         GOTO_TRAP(TraceTrap);
+      }
    }
 }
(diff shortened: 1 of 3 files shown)


SteveFosdick (approved):
This looks reasonable.  The NS32016 emulation is from PiTubeDirect so it may be worth mentioning to David Banks (on here, Hoglet on Stardot) in case he wants to include this.

